PR #1, opened 2019-03-26
LT-894: I want to submit my licence application when it is complete

jfaracik:
- Initial upload
- Hub
- You can:
-- Create an application for an export licence from the LITE exporter dashboard
-- See a summary of my draft and submitted applications on my exporter dashboard
-- Submit my licence application when it is complete
-- Cancel the application I am currently working on


applications/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path
+
+from . import views
+
+app_name = 'applications'
+urlpatterns = [
+    # ex: /
```
anirbanlahiri-dit: Is there any particular reason for this comment? I not please remove them

applications/urls.py
```diff
@@ -0,0 +1,11 @@
...
+    path('', views.index, name='applications'),
+  	# ex: /
```
anirbanlahiri-dit: same here

conf/settings.py
```diff
@@ -0,0 +1,156 @@
...
+from environ import Env
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+ENV_FILE = os.path.join(BASE_DIR, '.env')
+if os.path.exists(ENV_FILE):
+    Env.read_env(ENV_FILE)
+
+env = Env(
```
anirbanlahiri-dit: This has been fixed in lite_internal_frontend now, please take a look

conf/settings.py
```diff
@@ -0,0 +1,156 @@
...
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '=(kaa@ypr5v!x(s=9^f8)o!k#84f_1v@iz31+cq_)8--@kws4b'
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = env('DEBUG')
+
+ALLOWED_HOSTS = json.loads(env('ALLOWED_HOSTS')) if env('ALLOWED_HOSTS') else []
```
anirbanlahiri-dit: This has been sorted in lite_internal_frontend. Please take a look.

conf/settings.py
```diff
@@ -0,0 +1,156 @@
...
+
+STATICFILES_FINDERS = (
+    'django.contrib.staticfiles.finders.FileSystemFinder',
+    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
+    'sass_processor.finders.CssFinder',
+)
+
+SASS_PROCESSOR_ENABLED = True
+
+# Database
```
anirbanlahiri-dit: Please remove

form/models.py
```diff
@@ -0,0 +1,52 @@
+import uuid
+from enum import Enum
+
+
+class Section:
```
anirbanlahiri-dit: These classes should not be in models.py as they are not models

form/models.py
```diff
@@ -0,0 +1,52 @@
+import uuid
```
anirbanlahiri-dit: Please rename this file as helpers.py since it contain classes

form/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+class Section:
+    def __init__(self, title, description, forms):
+        self.id = uuid.uuid1()
+        self.title = title
+        self.description = description
+        self.forms = forms
+
+
+class Form:
```
anirbanlahiri-dit: The App is called form and here we have class Form, which might lead to confusion later, can we rename it to something more intuitive

libraries/jsondate/__init__.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2012 Rick Harris
```
anirbanlahiri-dit: Lets discuss if we really need this library and the method for selecting library

assets/css/styles.scss
```diff
@@ -0,0 +1,47 @@
+// GOV.UK Design System
+@import "../../node_modules/govuk-frontend/all";
+
+// LITE Shared Assets
+@import "../../node_modules/lite-shared-assets/scss/all";
+
+// Local Assets
```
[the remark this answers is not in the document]
jfaracik: We need the SCSS imports.

jfaracik:
> Do we still need the package.json and package-lock.json ? We are not using nom install, so can we remove it please?

We'll be using NPM in the future, so while we don't strictly need it, it'll prove very helpful in the future if we just left it.

form/models.py
jfaracik: I'll resolve this when we eventually move it to a separate library.

libraries/jsondate/__init__.py
jfaracik: This library is used for converting strings in JSON to the appropriate Python data type. Used for showing friendly dates and times on the front end.

ui_automation/tests/draft_application.py
```diff
@@ -0,0 +1,69 @@
...
+
+class DraftTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # get the path of ChromeDriverServer
+        project_root = os.path.dirname(os.path.abspath(__file__))
+        base_dir = os.path.dirname(project_root)
+        print("dir:" + base_dir)
+
+        chrome_driver_path = base_dir + "/drivers/chromedriver"
```
anirbanlahiri-dit: The chrome_driver_path should ideally not be hardcoded

ui_automation/tests/draft_application.py
```diff
@@ -0,0 +1,69 @@
...
+        # create a new Chrome session
+        cls.driver = webdriver.Chrome(chrome_driver_path)
+        cls.driver.implicitly_wait(30)
+        cls.driver.maximize_window()
+
+        # navigate to the application home page
+        cls.driver.get("https://lite-exporter-frontend-dev.london.cloudapps.digital/")
+
+    def test_start_draft_application(self):
```
anirbanlahiri-dit: This test works well. Would be useful to have a few tests for checking what's happens if incorrect information is entered on the browser, or fields are left empty.

ui_automation/tests/draft_application.py
anirbanlahiri-dit: Since the chrome driver is only about 15MB, would it be worthwhile to just put it in the repo? However, both Linux and Mac versions would potentially need to be kept

wisdome123: hardcoded path removed
no need to put it in the repo, versions change all the time

PR #5, opened 2019-04-03
JSON/SCSS Refactor

jfaracik:
- Removed JSON Date library
- Now using a custom tag to convert strings in templates to dates
- Now using `response.json()` instead of `json.loads(response.text)` (simpler)
- Now compress CSS for faster load times
- Increased width of full_width pages, fixed full_width margin issue


applications/views.py
```diff
@@ -10,18 +8,17 @@ def index(request):
...
 
 def application(request, id):
     response = requests.get(env("LITE_API_URL") + '/applications/' + str(id) + '/')
-    data = json.loads(response.text)
 
     context = {
-        'data': data,
-        'title': data.get("application").get("name"),
+        'data': response.json(),
+        'title': response.json().get("application").get("name"),
```
jakub-kozlowski: Do we have to check for `null`s (whether or not they exist in Python) ?

templates/applications/application.html
```diff
@@ -21,9 +21,9 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
 			<p class="govuk-label">
 				<a class="hoverable-link tooltip">
 					<span class="tooltip-content" style="min-width: 150px;">
-						{{ data.application.last_modified_at }}
+						{{ data.application.last_modified_at|str_date }}
 					</span>
-					{{ data.application.last_modified_at|naturaltime }}
+					{{ data.application.last_modified_at|str_date|naturaltime }}
 				</a>
```
jakub-kozlowski: Separate filters with spaces for better readability? (eg. `str_date | naturaltime`)

applications/views.py
jfaracik: It's doable, but I figure we've got bigger issues if the API is returning null.

templates/applications/application.html
jfaracik: No spaces seems to be convention https://docs.djangoproject.com/en/2.1/howto/custom-template-tags/

core/builtins/custom_tags.py
```diff
@@ -1,10 +1,20 @@
 from django import template
+from django.template.defaultfilters import stringfilter
 
 from conf import strings
+import datetime
 
 register = template.Library()
 
+ISO8601_FMT = '%Y-%m-%dT%H:%M:%SZ'
```
anirbanlahiri-dit: Can we put this in a standard confirmation library that we use across the project since it is standard

core/builtins/custom_tags.py
```diff
@@ -1,10 +1,20 @@
...
+
 
 @register.simple_tag
 def get_string(value):
     return strings.constants.get(value, 'COULDN\'T FIND STRING')
```
anirbanlahiri-dit: returning a string is not desirable

anirbanlahiri-dit:
OK to merge to master as a "fix: JSON date library fixed and SCSS refactored"

PR #19, opened 2019-04-25
refactor: docker support

mh-hmcts:


Procfile
```diff
@@ -1 +0,0 @@
-web: ./manage.py migrate && ./manage.py runserver 0.0.0.0:$PORT
```
jakub-kozlowski: We still need Procfile for gov paas.

sample.env
```diff
@@ -1,4 +1,5 @@
 DEBUG=True
-ALLOWED_HOSTS='["localhost"]'
-LITE_API_URL='http://127.0.0.1:8000'
-DATABASE_URL='postgres://postgres:password@localhost:5432/postgres'
+ALLOWED_HOSTS='["*"]'
+LITE_API_URL='api:8100'
```
jakub-kozlowski: does that need http in front?

mh-hmcts: nope seems to work

PR #22, opened 2019-04-26
Lt 1073 e2e

wisdome123:


ui_automation_tests/tests/00_setup_test.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.fixture(scope="function")
+def open_internal_hub(driver, internal_url):
+    # navigate to the application home page
+    driver.get(internal_url)
+    # driver.maximize_window()
+    print(driver.current_url)
+
+
+def test_new_organisation_setup(driver, open_internal_hub):
+    print("Setting up new organisation")
```
anirbanlahiri-dit: Would it be better to log to the console rather than print?

PR #73, opened 2019-07-08
Lt 1006 raise a clc

5pence:


goods/views.py
```diff
@@ -21,14 +24,57 @@ def get(self, request, **kwargs):
...
 
-        if status_code == 400:
-            return form_page(request, forms.form, request.POST, errors=data['errors'])
+    def get(self, request, **kwargs):
+        return form_page(request, self.main_form)
+
+    def post(self, request):
+        data = request.POST.copy()
+
+        # Logic for when we are the confirmation page
```
jfaracik: For when we are at?

goods/views.py
```diff
@@ -21,14 +24,57 @@ def get(self, request, **kwargs):
...
+        data['validate_only'] = False
+
+        if 'clc_query_confirmation' in data:
+            if data['is_good_controlled'] == 'unsure' and data['clc_query_confirmation'] == 'yes':
+                data['are_you_sure'] = True
+                print('==============================')
```
jfaracik: print statements

PR #83, opened 2019-07-24
feat: LT-758 - Exporter SSO

cstobart:


conf/views.py
```diff
@@ -0,0 +1,60 @@
...
+from raven.contrib.django.raven_compat.models import client
+from auth.utils import get_client, AUTHORISATION_URL, TOKEN_URL, \
+    TOKEN_SESSION_KEY
+
+
+class AuthView(RedirectView):
+    permanent = False
+
+    def get_redirect_url(self, *args, **kwargs):
+
```
andytaylorkainos: remove blank line

conf/views.py
```diff
@@ -0,0 +1,60 @@
...
+        authorization_url, state = get_client(self.request).authorization_url(AUTHORISATION_URL)
+
+        self.request.session[TOKEN_SESSION_KEY + '_oauth_state'] = state
+
+        return authorization_url
+
+
+class AuthCallbackView(View):
+    def get(self, request, *args, **kwargs):
+
```
andytaylorkainos: blank line

docker.env
```diff
@@ -8,3 +8,8 @@ GOOGLE_ANALYTICS_KEY=
 PYTHONUNBUFFERED=1
 TEST_SSO_EMAIL=
 TEST_SSO_PASSWORD=
+
+# Auth
+AUTHBROKER_CLIENT_ID=
+AUTHBROKER_CLIENT_SECRET=
+AUTHBROKER_URL=
```
andytaylorkainos: newline

local.env
```diff
@@ -1,8 +1,14 @@
...
+GOOGLE_ANALYTICS_KEY=
 PORT=9000
-FEEDBACK_URL='mailto:'
+FEEDBACK_URL=mailto:
+
+
+# Auth
+AUTHBROKER_CLIENT_ID=
+AUTHBROKER_CLIENT_SECRET=
+AUTHBROKER_URL=
```
andytaylorkainos: newline

core/models.py
```diff
@@ -43,6 +43,8 @@ class User(AbstractUser):
     email = models.EmailField(default=None, blank=True, unique=True)
     first_name = models.CharField(max_length=30, blank=True, default=None, null=True)
     last_name = models.CharField(max_length=150, blank=True, default=None, null=True)
+    user_token = models.TextField(default=None, blank=True, null=True)
+    backend_id = models.TextField(default=None, blank=True, null=True)
```
andytaylorkainos: should this variable not be named a little more obviously - something like `lite_api_id`?

Pipfile
```diff
@@ -25,6 +25,8 @@ psycopg2-binary = "==2.7.7"
 django-inline-svg = "*"
 pytest-bdd = "*"
 pyautogui = "*"
+raven = "*"
+requests-oauthlib = "*"
```
andytaylorkainos: Can we not set the explicit working version? Everytime the pipfile.lock is regenerated it'll grab the latest version which `may` break something

cstobart: I would prefer the security of the up to date packages and we can fix if they update

core/models.py
lukesa: `lite_api_user_id`

mh-hmcts: even better if we could rework the endpoint to accept the token for this particular case, as the token identifies the logged in user to the backend api service

PR #96, opened 2019-08-19
feat: LT-1175 - Administer applications for multiple companies

jfaracik:
* Added 'switch your organisation' page, enabling consultants and users to switch between their various organisations
* Added organisation ID to the headers
* Cleanup around the place (eg using the shiny new _lite-tag_ instead of the dirty old _lite-info-label_)


efahy:
The logout button takes me to a page that doesn't exist on Great

efahy:
After "Logging out" of the external site, I would expect to need to login to Great next time I hit LITE but at the moment on the 2nd visit I get taken straight to the LITE select org page.  For me that means the log out button is not living up to its hype

efahy:
You might consider putting a "Back" link on the Select Org page in situations where they have previously selected an org, i.e. whenever they aren't coming directly in from Great having just authenticated

auth/views.py
```diff
@@ -71,23 +72,32 @@ def get(self, request, *args, **kwargs):
...
         user.last_name = response['last_name']
         user.lite_api_user_id = response['lite_api_user_id']
         user.save()
+
         if user is not None:
             login(request, user)
 
+            user_dict, _ = get_user(request)
+
+            if len(user_dict['user']['organisations']) > 1:
```
efahy: Please also cater gracefully for the edge case where a user doesn't have any org

templates/apply_for_a_licence/external_locations/index.html
```diff
@@ -9,44 +9,36 @@
...
 		</div>
 	</div>
 
+	{% if data.external_locations|length %}
+		{% for external_location in data.external_locations %}
+			<div class="lite-card">
+				<!--<a href="#" class="lite-card-delete">
+					{% svg 'close' %}
+					<span class="govuk-visually-hidden">Remove location</span>
+				</a>-->
```
efahy: Remove commented out code if its not needed

templates/layouts/base.html
```diff
@@ -65,19 +65,19 @@
...
+                    <div class="lite-user-menu-divider"></div>
+
+                    <a class="lite-user-menu-button" href="{% url 'users:profile' %}">
+                        {{ user.first_name }}
+                        <div class="lite-user-menu-button--user">
+                            {% svg 'user' %}
+                        </div>
+                    </a>
+                </div>
+            {% endif %}
```
efahy: Indenting wrong?

templates/users/index.html
```diff
@@ -14,46 +14,47 @@
...
+		<div class="lite-tabs--fade">
+			<div class="lite-tabs">
+				<a href="{% url 'users:users' %}" class="lite-tabs__tab selected">Members</a>
+				<a href="{% url 'sites:sites' %}" class="lite-tabs__tab">Sites</a>
+			</div>
 		</div>
 		<div class="lite-app-bar__controls">
 			<a href="{% url 'users:add' %}" class="govuk-button" draggable="false" role="button">
-				Add a user
+				Add a member
```
efahy: Please run past content

PR #108, opened 2019-08-29
Dev2

phlptt:


ui_automation_tests/conftest.py
```diff
@@ -39,6 +39,9 @@ def pytest_addoption(parser):
     if env == 'local':
         parser.addoption("--exporter_url", action="store", default="http://localhost:9000", help="url")
         parser.addoption("--lite_api_url", action="store", default="http://localhost:8100", help="url")
+    elif env == 'dev2':
```
jfaracik: Would it not be best just to have exporter_url and lite_api_url in the env instead of this if statement?

phlptt: would have to change jenkins config too

PR #116, opened 2019-09-04
feat: LT-1007 - Raise an end user advisory query

jfaracik:
* Added form for applying for an end user advisory query
* Added list of queries


core/views.py
```diff
@@ -36,6 +36,8 @@ def get(self, request, **kwargs):
                          reverse_lazy('applications:applications')),
                     Tile('Goods', generate_notification_string(num_clc_notifications),
                          reverse_lazy('goods:goods')),
+                    Tile('End Users', '',
```
cstobart: Is this the title we're going with as we don't have the concept of an address book? Maybe something more like "End User Advisories"?

cstobart: Is this the title we're going with as we don't have the concept of an address book? Maybe something more like "End User Advisories"?

end_users/forms.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def apply_for_an_end_user_advisory_form():
+    return FormGroup([
+        Form(title='Check if someone is eligible to import your goods',
+             questions=[
+                 RadioButtons(title='How would you describe this end user?',
+                              name='end_user.type',
+                              options=[
+                                  Option('government', 'A Government Organisation'),
```
cstobart: Could these options be stored in a list that could be pulled in?

templates/end_users/index.html
```diff
@@ -0,0 +1,59 @@
...
+			<a href="{% url 'end_users:apply' %}" class="govuk-button" draggable="false" role="button">
+				Apply for an end user advisory
+			</a>
+		</div>
+	</div>
+
+	{% if end_users|length %}
+		<table class="govuk-table">
+			<thead class="govuk-table__head">
+				<tr class="govuk-table__row">
```
cstobart: Table generator?

core/builtins/custom_tags.py
```diff
@@ -84,3 +84,9 @@ def insert_str(string, str_to_insert, string_index):
         value = insert_str(value, span_end, index + len(span) + len(term))
 
     return value
+
+
+@register.filter()
+def reference_code(value):
+    value = str(value)
+    return value[:5] + '-' + value[5:]
```
cstobart: Is this fixed to 10 digits?

core/views.py
```diff
@@ -36,6 +36,8 @@ def get(self, request, **kwargs):
                          reverse_lazy('applications:applications')),
                     Tile('Goods', generate_notification_string(num_clc_notifications),
                          reverse_lazy('goods:goods')),
+                    Tile('End Users Advisories', '',
```
cstobart: `Users` or `User` ?

end_users/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class EndUsersConfig(AppConfig):
+    name = 'end_users'
```
cstobart: Necessary?

end_users/views.py
```diff
@@ -0,0 +1,50 @@
...
+    def get(self, request, **kwargs):
+        return form_page(request, self.forms.forms[0])
+
+    def post(self, request, **kwargs):
+        response, data = submit_paged_form(request, self.forms, post_end_user_advisories)
+
+        if response:
+            return response
+
+        return success_page(request,
```
cstobart: Could this be a method calling a specific success page from a file containing many of these configs e.g.:

return end_user_adviosry_success_page()

PR #121, opened 2019-09-11
feat: LT-1007, LT-1138 - Create and respond to queries

jfaracik:
* Can now create End User Advisory Queries
* Responses from Gov can now be seen for CLC Queries
* New Verified status for goods


efahy:
Alignment of column text is off on the exporter goods list page - the name Description text is lower than the text in the other columns

ui_automation_tests/features/end_user_advisory_queries.feature
```diff
@@ -0,0 +1,12 @@
+@end_user_advisory_queries @all
+Feature:
```
phlptt: add a feature name here and the user story from jira

phlptt:
needs tests for 1138

PR #129, opened 2019-09-18
feat: LT-1483 - Copy end user advisory queries

jfaracik:
* Added the ability to copy end user advisory queries
* Added the declaration page at the start of the form group


end_users/views.py
```diff
@@ -20,6 +22,41 @@ def get(self, request, **kwargs):
...
+class CopyAdvisory(TemplateView):
+
+    forms = None
+    data = None
+
+    def dispatch(self, request, *args, **kwargs):
+        self.forms = copy_end_user_advisory_form()
+        query = get_end_user_advisory(request, str(kwargs['pk']))
+
+        self.forms.forms[0].questions.append(HiddenField('end_user.type', query['end_user']['type']['key']))
```
andytaylorkainos: Can you add a comment to explain what form is expected in `forms[0]`?

end_users/urls.py
```diff
@@ -7,6 +7,10 @@
 urlpatterns = [
     # ex: /end-users/
     path('', views.EndUsersList.as_view(), name='end_users'),
+
+    # ex: /<int:pk>/copy/
+    path('<int:pk>/copy/', views.CopyAdvisory.as_view(), name='copy'),
```
andytaylorkainos: can we rename the endpoint? it looks like you're copying an `end_user` instead of the advisory data

PR #134, opened 2019-09-23
Add lint

lukesa:


goods/views.py
```diff
@@ -259,18 +259,19 @@ def post(self, request, **kwargs):
...
-            'page': 'goods/modals/delete_document.html',
-        }
+        # TODO: should we remove context variable below?
+        # context = {
+        #     'title': 'Are you sure you want to delete this file?',
+        #     'description': document['name'],
+        #     'good': good['good'],
+        #     'document': document,
+        #     'page': 'goods/modals/delete_document.html',
+        # }
```
andytaylorkainos: have you verified this TODO? if we don't need it, delete it pls :)

ui_automation_tests/helpers/seed_data.py
```diff
@@ -101,11 +101,12 @@ class SeedData:
             'part_number': '1234',
             'validate_only': False,
         },
-        'gov_user': {
+        # TODO: Investigate duplication: gov_user / export_user
```
andytaylorkainos: again, if there's duplication - remove it please

goods/views.py
lukesa: this is delete feature for potential future story

PR #135, opened 2019-09-23
feat: LT-1486 - Add Additional Details to End User Advisory

cstobart:


end_users/forms.py
```diff
@@ -106,7 +135,7 @@ def copy_end_user_advisory_form():
                  HiddenField('validate_only', False),
              ],
              default_button_name='Submit'),
```
jfaracik: And here

PR #145, opened 2019-09-26
LT-1499-bug-end-user-done

BenAndrew-Kainos:


apply_for_a_licence/views/common.py
```diff
@@ -79,9 +86,12 @@ def get(self, request, **kwargs):
             'external_locations': external_locations['external_locations'],
             'ultimate_end_users': ultimate_end_users['ultimate_end_users'],
             'ultimate_end_users_required': ultimate_end_users_required,
+            'ultimate_end_users_documents_complete': 
+                check_all_parties_have_documents(ultimate_end_users['ultimate_end_users']),
```
stephanieboyd: minor - but can we change this to `a_document` - as the parties can only have a single document

BenAndrew-Kainos: will do

PR #167, opened 2019-10-04
refactor: Added service name placeholder

jfaracik:
* Added service name placeholder
* Corrected breadcrumbs
* Removed unused file


templates/layouts/base.html
```diff
@@ -64,18 +64,23 @@
...
-                    </a>
+					<a href="{% url 'authbroker_client:logout' %}" class="lite-user-menu-button">Sign out</a>
+					<div class="lite-user-menu-divider"></div>
+
+					<a class="lite-user-menu-button" href="{% url 'users:profile' %}">
+						{{ user.first_name }}
+						<div class="lite-user-menu-button--user">
+							{% svg 'user' %}
+						</div>
+					</a>
```
markm5A8Z9F43913A: Why so far right?

craiga: Unnecessary extra indentation.

conf/client.py
```diff
@@ -14,23 +14,23 @@ def get(request, appended_address):
...
+    return requests.put(env('LITE_API_URL') + appended_address,
                         json=json,
                         headers={'EXPORTER-USER-TOKEN': str(request.user.user_token),
                                  'X-Correlation-Id': str(request.correlation),
                                  'ORGANISATION-ID': str(request.user.organisation)})
 
 
 def delete(request, appended_address):
-    return requests.delete(env("LITE_API_URL") + appended_address,
+    return requests.delete(env('LITE_API_URL') + appended_address,
```
craiga: FWIW, `black` is going to change these back.

conf/context_processors.py
```diff
@@ -1,6 +1,9 @@
 import os
 
+from conf import settings
+
 
 def export_vars(request):
-    data = {'ENVIRONMENT_VARIABLES': dict(os.environ.items())}
+    data = {'ENVIRONMENT_VARIABLES': dict(os.environ.items()),
+            'SERVICE_NAME': settings.SERVICE_NAME}
```
craiga: Woah. I know it's got nothing to do with your change, but putting all environment variables into context seems like a massive security risk.

templates/layouts/base.html
```diff
@@ -50,7 +50,7 @@
 		<div class="govuk-header__container govuk-width-container">
 
 			<div class="govuk-header__logo">
-				<a href="/" class="govuk-header__link govuk-header__link--homepage">
+				<a href="https://www.gov.uk" class="govuk-header__link govuk-header__link--homepage">
```
andytaylorkainos: very picky, but can we make this a constant? (like you did with the service name)

templates/layouts/base.html
jfaracik: It isn't - Github is just be dodgy

templates/core/hub.html
```diff
@@ -18,6 +18,13 @@
 {% endblock %}
 
 {% block body %}
+
+	<style media="screen">
+		.govuk-header__content {
+			display: none;
+		}
+	</style>
```
craiga: Why inline styles? Do we not have a separate CSS file this can go in?

templates/layouts/base.html
jfaracik: Could do! - Consider it _done_

conf/context_processors.py
jfaracik: You're right! I'll resolve that

templates/core/hub.html
jfaracik: Inline just to hide the govuk heading for this one page

templates/layouts/base.html
jfaracik: This doesn't show in atom or PyCharm for me :/

conf/client.py
jfaracik: no longer want Black - single quotes for life :(

conf/context_processors.py
```diff
@@ -1,6 +1,11 @@
...
 
 
 def export_vars(request):
-    data = {'ENVIRONMENT_VARIABLES': dict(os.environ.items())}
+    data = {
+        'SERVICE_NAME': 'LITE',
+        'GOV_UK_URL': 'https://www.gov.uk',
+        'FEEDBACK_URL': env('FEEDBACK_URL'),
+        'GOOGLE_ANALYTICS_KEY': env('GOOGLE_ANALYTICS_KEY'),
+    }
```
craiga: 👍

templates/core/hub.html
craiga: It's a slippery slope—inline CSS to "just do this one thing" will quickly become messy and unmanageable. I won't hold this PR up because of it, but if there is somewhere else this could go it should go there.

conf/context_processors.py
```diff
@@ -1,6 +1,11 @@
-import os
+from conf.settings import env
```
craiga: Pulling in this instance of `environ.Env` might cause problems down the road. I could imagine someone reconfiguring this object in `conf.settings` and unintentionally breaking this.

Would suggest reading these values into `django.conf.settings` and using from there.

templates/layouts/base.html
craiga: Looks like the whole block has extra indentation now? 🤷‍♂

I'm assuming the only change here has to do with the changes to the context processor.

PR #173, opened 2019-10-09
refactor: LT-998 integrate /drafts/ into /applications/

andytaylorkainos:


templates/applications/tabs/edit-application.html
```diff
@@ -0,0 +1 @@
+<div>Hello world!</div>
```
markm5A8Z9F43913A: Probably should delete this :D

templates/drafts/index.html
```diff
@@ -25,7 +25,7 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
 		</div>
 	</div>
 
-	{% if data.drafts|length %}
+	{% if data.applications|length %}
```
craiga: I _think_ `{% if data.applications %}` would work just as well here … a zero-length list is false-ish.

applications/views.py
```diff
@@ -32,6 +33,12 @@ def get(self, request, **kwargs):
                                                                                 'type': 'case-notes'}))
 
 
+class ApplicationEdit(TemplateView):
+
+    def get(self, request, **kwargs):
+        return get_licence_overview(request, kwargs)
```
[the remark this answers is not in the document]
andytaylorkainos: Apologies Craig, this code should not have made it into the PR (it has since been removed)

templates/drafts/index.html
andytaylorkainos: Think this might have been done originally for a reason, going to change it and watch e2e :)

PR #180, opened 2019-10-11
Refactor shared fixtures

BenAndrew-Kainos:


ui_automation_tests/conftest.py
```diff
@@ -4,16 +4,19 @@
...
+from ui_automation_tests.fixtures.add_clc_query import add_clc_query  # noqa
+from ui_automation_tests.fixtures.add_end_user_advisory import add_end_user_advisory  # noqa
+from ui_automation_tests.fixtures.internal_ecju_query import internal_ecju_query, internal_ecju_query_end_user_advisory  # noqa
+from ui_automation_tests.fixtures.sso_sign_in import sso_sign_in  # noqa
+from ui_automation_tests.fixtures.internal_case_note import internal_case_note, internal_case_note_end_user_advisory  # noqa
+
+from ui_automation_tests.shared.fixtures.apply_for_application import apply_for_standard_application, add_an_ecju_query  # noqa
+from ui_automation_tests.shared.fixtures.driver import driver  # noqa
+from ui_automation_tests.shared.fixtures.core import context, invalid_username, exporter_info, internal_info, s3_key, seed_data_config  # noqa
+from ui_automation_tests.shared.fixtures.urls import exporter_url, api_url  # noqa
```
craiga: Why all the `#noqa` here?

Couldn't we do something like this instead?

```python
from ui_automation_tests.shared.fixtures.apply_for_application import (
    apply_for_standard_application,
    add_an_ecju_query,
)
```

BenAndrew-Kainos: Unfortunately CircleCI doesn't detect these imports as being used because there aren't directly referenced from within conftest, but need to be here for the test suite to be able to access them. #noqa tells circleCI to ignore the imports

PR #188, opened 2019-10-22
feat: LT-998 - Edit application

jfaracik:
* Overhauled Overview page - now closer to prototype
* Now possible to edit applications (minor and major edits)
* Cleaned up a lot of service functions
* Cleaned up old views


applications/forms.py
```diff
@@ -32,5 +33,23 @@ def ecju_query_respond_confirmation_form(edit_response_url):
...
+                                     Option(key='minor',
+                                            value='Change a site, or delete a good, third party or country',
+                                            description='This won\'t impact the time it takes to reach a decision.'),
+                                     Option(key='major',
+                                            value='Something else',
+                                            description='This will impact the time it takes to reach a decision.')
+                                 ])],
+                back_link=BackLink('Back to application', reverse_lazy('applications:application-detail',
+                                                                       kwargs={'pk': application_id,
+                                                                               'type': 'ecju-queries'})),
```
Mark-Spiers-kainos: will this always be ecju-queries?

applications/views.py
```diff
@@ -1,37 +1,97 @@
...
+        return form_page(request, edit_type_form(application_id))
+
+    def post(self, request, **kwargs):
+        if request.POST.get('edit-type') == 'major':
+            data, status_code = submit_application(request, str(kwargs['pk']), 'applicant_editing')
+
+            if status_code != HTTPStatus.OK:
+                return form_page(request, edit_type_form(str(kwargs['pk'])), errors=data)
+
+        return redirect(reverse_lazy('applications:application_edit_overview', kwargs={'pk': str(kwargs['pk'])}))
```
Mark-Spiers-kainos: perhaps consider making str(kwargs['pk']) a variable at start of method.

ui_automation_tests/features/case_notes.feature
```diff
@@ -20,11 +20,21 @@ Feature: I want to add a note to an application and view notes
...
-    And an application exists
+    And a standard application exists
     When I click on applications
     And I click on application previously created
     And I enter "the maximum limit" for case note
     Then case note warning is "You have 0 characters remaining"
     When I enter "the maximum limit plus 1" for case note
     Then case note warning is "You have 1 character too many"
     And post note is disabled
+
```
andytaylorkainos: This looks like a merge conflict issue (compare with master)

applications/urls.py
```diff
@@ -1,15 +1,133 @@
...
-    # ex: /applications/43a88949-5db9-4334-b0cc-044e91827451/case-notes/
-    path('<uuid:pk>/<str:type>/', views.ApplicationDetail.as_view(), name='application-detail'),
-    # ex: /applications/43a88949-5db9-4334-b0cc-044e91827451/
-    path('<uuid:pk>/ecju-queries/<uuid:query_pk>/', views.RespondToQuery.as_view(), name='respond_to_query'),
+    path('', common.ApplicationsList.as_view(), name='applications'),
+    # ex: /applications/<uuid:pk>/
+    path('<uuid:pk>/', common.ApplicationDetailEmpty.as_view(), name='application'),
+    # ex: /<uuid:pk>/delete/
+    path('<uuid:pk>/delete/', common.DeleteApplication.as_view(), name='delete'),
+    # ex: /applications/<uuid:pk>/edit/overview
```
Mark-Spiers-kainos: 'overview' not in the url

applications/views/locations.py
```diff
@@ -1,27 +1,38 @@
...
 class Location(TemplateView):
     def get(self, request, **kwargs):
-        draft_id = str(kwargs['pk'])
-        data, _ = get_external_locations_on_draft(request, draft_id)
-        if data['external_locations']:
+        application_id = str(kwargs['pk'])
+        application = get_application(request, application_id)
+        external_locations, _ = get_external_locations_on_draft(request, application_id)
+
+        print(application['status'])
```
Mark-Spiers-kainos: remove please

templates/applications/tabs/edit-application.html
```diff
@@ -0,0 +1 @@
+<div>Hello world!</div>
```
Mark-Spiers-kainos: is this file needed?

jfaracik: Nice spot!

applications/libraries/get_licence_overview.py
```diff
@@ -0,0 +1,82 @@
+from django.shortcuts import render
+
+from applications.services import get_additional_documents, get_ultimate_end_users, get_third_parties, \
+    get_application_goods, get_end_user_document, get_consignee_document, get_application_countries, \
+    get_application_goods_types
```
craiga: At this point, it might be easier to just `from applications import services` here and call `services.get_additional_documents`, etc.

applications/forms/common.py
```diff
@@ -32,5 +32,23 @@ def ecju_query_respond_confirmation_form(edit_response_url):
                         no_label='No, change my response',
                         back_link_text='Back to edit response',
                         back_url=edit_response_url,
-                        submit_button_text="Continue"
-                        )
+                        submit_button_text='Continue')
+
+
+def edit_type_form(application_id):
+    return Form(title='How do you want to change your application?',
```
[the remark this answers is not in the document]
a-bek: Darth-Vader-1200x676

applications/libraries/get_licence_overview.py
andytaylorkainos: 👍

core/builtins/custom_tags.py
```diff
@@ -112,3 +112,38 @@ def times(number):
...
+    else:
+        return mark_safe('<span class="lite-hint">N/A</span>')  # nosec
+
+
+@register.filter()
+def friendly_boolean(boolean):
+    """
+    Returns 'Yes' if a boolean is equal to True, else 'No'
+    """
+    if boolean is True or boolean == 'true' or boolean == 'True':
```
a-bek: Could just do `boolean.lower() == 'true'` which will cover all capitalisation permutations.

core/builtins/custom_tags.py
```diff
@@ -112,3 +112,38 @@ def times(number):
...
+        return 'Yes'
+    else:
+        return 'No'
+
+
+@register.filter()
+def pluralise_unit(unit, value):
```
a-bek: Where does `(s)` come from?

applications/forms/common.py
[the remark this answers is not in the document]
jfaracik: giphy-2

core/services.py
```diff
@@ -18,19 +20,12 @@ def get_units(request):
...
-
-        for country in data.json().get('countries'):
-            converted_units.append(
-                Option(country.get('id'), country.get('name'))
-            )
+        return [Option(x['id'], x['name']) for x in data]
 
-        return converted_units
-
-    return data.json(), data.status_code
```
a-bek: There are a lot of `get_*()` functions around. Some return a tuple, some dont. Should this be standardised? Maybe we could create like a custom nametuple response object that can be used everywhere.

```
# defined in another module somewhere
Response = namedtuple('Response', ['data', 'status'])
```
```
from ... import Response

def get_countries(request....) -> Response:
     ...
     return Response(data=data.json(), status=data.status_code)
```

Just an idea, but doesn't need to be done in this PR.

andytaylorkainos: I think what needs to be done is a generalised `get` helper - one which optionally returns the status_code (so we can handle errors in the frontend)

core/builtins/custom_tags.py
jfaracik: Good suggestion

core/builtins/custom_tags.py
jfaracik: From the API https://github.com/uktrade/lite-api/blob/master/static/units/enums.py

core/services.py
jfaracik: There is some rough documentation on this on confluence. I don't personally like get_* returning status_code as get_* implies that it'll return that object and that object only. Worthy of more discussion though absolutely.

It'll take some time to standardise across LITE but we'll get there

core/builtins/custom_tags.py
jfaracik: I've added a docstring to this function to help clear up confusion.

core/builtins/custom_tags.py
andytaylorkainos: This doesnt work because this method expects either a bool or a string.
If it's passed a bool, it'll break because `False.lower()` can't be done
```
  File "/Users/andrewta/dit/lite-exporter-frontend/core/builtins/custom_tags.py", line 133, in friendly_boolean
    if boolean is True or boolean.lower() == 'true':
AttributeError: 'bool' object has no attribute 'lower'
```

andytaylorkainos: `if boolean is True or str(boolean).lower() == 'true':` seems hacky but there we go!

core/services.py
andytaylorkainos: We also need a discussion around what we should return from the API (should it always be a JSON string, with a status? - the frontends assume this)

PR #191, opened 2019-10-29
fix: Static assets with Whitenoise

jfaracik:
* Static assets now show on environments with DEBUG set to False.


Procfile
```diff
@@ -1 +1 @@
-web: ./manage.py migrate && gunicorn -c gconfig.py -b 0.0.0.0:$PORT conf.wsgi
+web: ./manage.py migrate && ./manage.py compilescss && ./manage.py collectstatic --ignore=*.scss,*.md,*.txt,*.json,LICENSE,license,CHANGES,changes && gunicorn -c gconfig.py -b 0.0.0.0:$PORT conf.wsgi
```
markm5A8Z9F43913A: Is it possible to split this into multiple lines to make it more readable?

conf/settings.py
```diff
@@ -56,14 +56,14 @@
 
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
+    'whitenoise.middleware.WhiteNoiseMiddleware',
```
markm5A8Z9F43913A: I'd really like this kind of stuff to be in alphabetical order. Something to keep in mind for the future.

craiga: The ordering of middleware is important. Alphabetical won't work well.

PR #196, opened 2019-11-05
enhance: Unit Test for Custom Tags

andytaylorkainos:


core/builtins/tests/test_custom_tags.py
```diff
@@ -0,0 +1,97 @@
...
+        span = '<span class="lite-highlight">'
+        span_end = '</span>'
+        actual_string = ct.highlight_text('this is a sassy string', 'a')
+        expected_string = 'this is ' + span + 'a' + span_end + ' s' + span + 'a' + span_end + 'ssy string'
+
+        self.assertEqual(actual_string, expected_string)
+
+    def test_strip_underscores(self):
+        expected_value = 'A long string'
+        actual_value = ct.strip_underscores('a_long_string')
```
markm5A8Z9F43913A: Why did it lose capitalisation here

andytaylorkainos: nono - it gained it

core/builtins/custom_tags.py
```diff
@@ -84,7 +84,10 @@ def insert_str(string, str_to_insert, string_index):
     span = '<span class="lite-highlight">'
     span_end = '</span>'
 
+    loop = 0
     for index in indexes:
+        index += loop*(len(span) + len(term) + len(span_end) - 1)
```
markm5A8Z9F43913A: I think a comment would be useful here.

core/builtins/tests/test_custom_tags.py
markm5A8Z9F43913A: Edited my question :)

andytaylorkainos: it's the expected functionality - the method is probably poorly named

PR #197, opened 2019-11-05
Bug: LT-1379 file size error

BenAndrew-Kainos:


applications/services.py
```diff
@@ -259,6 +259,10 @@ def add_document_data(request):
         return None, 'Multiple files attached'
 
     file = files[0]
+
+    if file.file_size > MAX_UPLOAD_SIZE:
+        return None, 'File too large'
```
jfaracik: Issue with this is that the file_size returned from a file can be overwritten by whatever is sending it, so it isn't necessarily correct

BenAndrew-Kainos: Yeah unfortunately I can't think of a better solution. Do you know of any?

andytaylorkainos: not an expert here but i've noticed there is `.size` and `.file_size) - seems most sources are using the former when comparing to `MAX_UPLOAD_SIZE`.

If in doubt, you could play with `request.headers['content-length']` (it will naturally be a bit larger than the file size alone)

PR #207, opened 2019-11-12
Adding support for LOCAL_LITE_API_URL environment variable.

craiga:
This does the same thing as [this pull request for lite-internal-frontend](https://github.com/uktrade/lite-internal-frontend/pull/195).

If you're running end-to-end tests from your local machine but the frontend and API are running in Docker containers, the LITE API URL will need to be configured twice—once for the frontend, and once for the end-to-end tests.

`LITE_API_URL` is the URL for the API as visible from the frontend. `LOCAL_LITE_API_URL` is the URL for the API as visible locally. If `LOCAL_LITE_API_URL` isn't set, `LITE_API_URL` will be used instead (i.e. current behaviour).


andytaylorkainos:
Im guessing this is for applications run in docker?

PR #220, opened 2019-11-21
Refactor: Case inheritance

Mark-Spiers-kainos:


applications/helpers/summaries.py
```diff
@@ -12,7 +12,7 @@ def draft_summary(draft):
             strings.APPLICATION_REFERENCE_NAME: draft["name"],
             strings.APPLICATION_TYPE: draft["application_type"]["value"],
             strings.APPLICATION_EXPORT_TYPE: draft["export_type"]["value"],
-            strings.APPLICATION_CREATED_AT: str_date(draft["created_at"]),
+            strings.APPLICATION_CREATED_AT: str_date(draft["created"]),
```
andytaylorkainos: howcome we changed it from `created_at`? 🤔

templates/applications/applications.html
```diff
@@ -73,7 +73,7 @@ <h1 class="govuk-heading-l">{% block title %}Applications{% endblock %}</h1>
 							</td>
 						{% endif %}
 						<td class="govuk-table__cell">
-							{{ application.last_modified_at|str_date }}
+							{{ application.modified|str_date }}
```
andytaylorkainos: same again, howcome we got rid of the `_at`?

applications/helpers/summaries.py
Mark-Spiers-kainos: We are using TimeStampedModel to inherit Case from, which is built into models_util. It includes code to auto update modified etc, and is an abstract model which we can inherit other classes from in the future.
Also modified actually updates now which is a bonus (for some changes in the system)

PR #221, opened 2019-11-22
Lt 1692 display org in exporter page header

efahy:


core/models.py
```diff
@@ -46,6 +46,7 @@ class User(AbstractUser):
     user_token = models.TextField(default=None, blank=True, null=True)
     lite_api_user_id = models.TextField(default=None, blank=True, null=True)
     organisation = models.UUIDField(default=uuid.uuid4, null=True, blank=True)
+    organisation_name = models.TextField(default=None, blank=True, null=True)
```
BenAndrew-Kainos: Do you want to allow this to be null?

efahy: Yes, because when they first log in through GREAT, there isn't a selected org so its none at that point

PR #227, opened 2019-11-28
fix: LT-1791 - Withdraw Application at Any Stage Except Finalised

cstobart:


templates/applications/application.html
```diff
@@ -12,7 +12,7 @@
             <h1 class="govuk-heading-l">{{ title }}</h1>
         </div>
         <div class="lite-app-bar__controls">
-			{% if application.status.value|lower not in read_only_statuses and application.application_type.key != 'hmrc_query' %}
+			{% if application.status.value|lower not in finalised_statuses and application.application_type.key != 'hmrc_query' %}
```
jfaracik: Should the edit button be hidden here too?

cstobart: good point

applications/views/common.py
```diff
@@ -176,6 +176,7 @@ def get(self, request, **kwargs):
                 context["open_queries"], context["closed_queries"] = get_application_ecju_queries(request, self.case_id)
 
         context["read_only_statuses"] = constants.READ_ONLY_STATUSES
+        context["finalised_statuses"] = constants.FINALISED_STATUSES
```
a-bek: Rather than send data, and do the check on the frontend, what do you think of removing logic from the templates.

`context['edit_button_visible'] = status not in constants.FINALISED_STATUSES`

then in  just do:

`{% if edit_button_visible %}`

cstobart: agreed

cstobart: done

templates/applications/application.html
```diff
@@ -12,10 +12,12 @@
             <h1 class="govuk-heading-l">{{ title }}</h1>
         </div>
         <div class="lite-app-bar__controls">
-			{% if application.status.value|lower not in read_only_statuses and application.application_type.key != 'hmrc_query' %}
+			{% if can_withdraw %}
```
markm5A8Z9F43913A: Tabs are used instead of spaces for indentation.

efahy:
Looks to me like you are one commit behind master for `lite-tests-common`.  Probably should re-attach to HEAD?

conf/constants.py
```diff
@@ -46,6 +46,7 @@
 
 # Case statuses
 READ_ONLY_STATUSES = ["finalised", "under final review", "under review", "withdrawn"]
+FINALISED_STATUSES = ["finalised", "withdrawn"]
```
stephanieboyd: Slight wording change needed. The statuses are 'terminal' as opposed to 'finalised' as finalised is a status. These constants are gonna go away shorty, but since this might be in first, the list should be:

`TERMINAL_CASE_STATUSES = ["closed", "deregistered", "finalised", "registered", "revoked", "surrendered", "withdrawn"]`

PR #228, opened 2019-11-28
fix: LT-1775 - Applications pagination

jfaracik:


templates/applications/applications.html
```diff
@@ -82,6 +82,17 @@ <h1 class="govuk-heading-l">{% block title %}Applications{% endblock %}</h1>
...
+
+		{% if applications.total_pages > 1 %}
+			<div class="lite-pagination__container">
+				{% for i in applications.total_pages|times %}
+					<a href="{% url 'applications:applications' %}?{{ params_str }}&page={{ i }}" class="lite-pagination__item {% if page == i %}lite-pagination__item--selected{% endif %}" id="page-{{ i }}">
+						{{ i }}
+					</a>
+				{% endfor %}
+			</div>
+		{% endif %}
```
markm5A8Z9F43913A: Looks like this code is being reused in multiple places. If you have time, it would be good to put everything in one place and just include it. Same thing on the internal side.

jfaracik: That's a very good suggestion! I'll see what I can do :)

jfaracik: I've added it - hope you like it :)

jfaracik:
> what's the seed_data change?

That was just pulling the latest submodules

core/views.py
```diff
@@ -13,7 +15,10 @@
 
 class Hub(TemplateView):
     def get(self, request, **kwargs):
-        user, _ = get_user(request)
+        try:
+            user, _ = get_user(request)
+        except JSONDecodeError:
+            return redirect("authbroker:login")
```
markm5A8Z9F43913A: How is failing to decode JSON directly related to not being able to log in?

templates/includes/pagination.html
```diff
@@ -0,0 +1,9 @@
+{% if data.total_pages > 1 %}
+	<div class="lite-pagination__container">
+		{% for i in data.total_pages|times %}
+			<a href="{{ CURRENT_PATH_WITHOUT_PARAMS }}?{{ params_str }}&page={{ i }}" class="lite-pagination__item {% if page == i %}lite-pagination__item--selected{% endif %}" id="page-{{ i }}">
```
markm5A8Z9F43913A: If there are no params apart from "page", you'll end up with `?&`

jfaracik: True, but is that really a big issue? I don't see it breaking things

core/views.py
jfaracik: This is more of a workaround for when the database is cleared. The API won't send JSON, the frontend will try to read the non existent JSON and fall over without a clear error (see delivery channel for an example). So now we handle that error and make them sign in again :)

templates/includes/pagination.html
markm5A8Z9F43913A: It's not correct syntax. You never know whose browser will be stupid or how Django will start handling it.

jfaracik: Done.

templates/end_users/end_users.html
```diff
@@ -35,7 +35,7 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
                             <a class="govuk-link govuk-link--no-visited-state app-icon-label" href="{% url 'end_users:end_user' pk=advisory.id %}">
 							    {{ advisory.end_user.name }}
                                 {% if advisory.id in notifications %}
-                                    <span class="app-icon-label__notification">{{ notifications|key_value:advisory.id }} <span class="govuk-visually-hidden"> notifications</span></span>
+                                    <span class="lite-notification-bubble">{{ notifications|key_value:advisory.id }} <span class="govuk-visually-hidden"> notifications</span></span>
```
andytaylorkainos: that's a beautiful string, prime candidate for lite_content :)

templates/goods/goods.html
```diff
@@ -48,7 +48,7 @@ <h2 class="govuk-heading-l">
 								<p>{{ good.description }}</p>
 								{% if good.status.key == 'verified' %}<span class="app-verified__symbol">{% svg 'verified' %}</span>{% endif %}
 								{% if good.query_id and good.query_id in notifications %}
-									<span class="app-icon-label__notification">{{ notifications|key_value:good.query_id }} <span class="govuk-visually-hidden"> notifications</span></span>
+									<span class="lite-notification-bubble">{{ notifications|key_value:good.query_id }} <span class="govuk-visually-hidden"> notifications</span></span>
```
andytaylorkainos: Is that another candidate i see? 👀

PR #232, opened 2019-12-02
Bug: LT-1764 Ultimate end user title string

BenAndrew-Kainos:


applications/views/third_parties.py
```diff
@@ -88,7 +89,7 @@ def init(self, request, **kwargs):
         self.object_pk = kwargs["pk"]
         application = get_application(request, self.object_pk)
         self.data = application["consignee"]
-        self.forms = new_consignee_forms(application)
+        self.forms = new_end_user_forms(application, ConsigneeForm)
```
andytaylorkainos: maybe rename the constructor method to `new_party_user_form` or something a bit more generic?

BenAndrew-Kainos: good idea. done

PR #234, opened 2019-12-03
chore: SSO refactor

jfaracik:
The purpose of this PR is to remove the amount of unused files, consolidate authentication into a single app, and make code more reusable.


conf/middleware.py
```diff
@@ -15,8 +17,8 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
-        if resolve(request.path).app_name != "authbroker_client" and not request.user.is_authenticated:
-            return redirect("authbroker_client:login")
+        if resolve(request.path).app_name != "auth" and not request.user.is_authenticated:
```
BenAndrew-Kainos: Could you store "auth" as a constant

ui_automation_tests/step_defs/test_users_steps.py
```diff
@@ -23,8 +20,6 @@ def add_user(driver):
         for x in range(3):
```
BenAndrew-Kainos: Who did this and why do we need it?

users/forms.py
```diff
@@ -24,20 +21,18 @@ def add_user_form(request):
     )
 
 
-def edit_user_form(request, user_id, super_user: bool):
+def edit_user_form(request, user):
     return Form(
-        title=strings.USER_EDIT_TITLE,
+        title="Change user role",
```
BenAndrew-Kainos: Move to strings

users/services.py
```diff
@@ -26,4 +26,7 @@ def update_user(request, pk, json):
 
 
 def is_super_user(user):
```
BenAndrew-Kainos: Could you change function usage instead to pass `user["user"]` where applicable

ui_automation_tests/step_defs/test_users_steps.py
jfaracik: I'm not too sure.

PR #235, opened 2019-12-03
Update third_party.py

anthony-lord:
Made changes to:
- H1
- radio button option
- button


BenAndrew-Kainos:
If it would help I can make these changes on your behalf 😄

PR #238, opened 2019-12-04
Update end_user.py

anthony-lord:
Removed 'final recipient' in H1s for the following journeys:
- add and end user
- and a consignee
- add a third party
- add an ultimate end user


anthony-lord:
Thanks Ben. I did ask in Slack where this content is, and Stephanie directed me here. We need to sort this out some how

PR #242, opened 2019-12-09
test: Add a implicit wait for the document to be uploaded

phlptt:


ui_automation_tests/step_defs/test_hmrc.py
```diff
@@ -38,3 +38,14 @@ def i_click_on_hmrc_set_end_user(driver):
...
     app = ApplicationOverviewPage(driver)
     app.click_hmrc_explain_your_reasoning()
+
+
+@when("I wait for document to upload")
+def wait_for_document(driver):
+    document_is_found = False
+    while not document_is_found:
+        if 'Processing' in driver.find_element_by_id('document').text:
+            driver.refresh()
```
andytaylorkainos: instead of refreshing every time, can we make it `sleep` for 10 seconds?

ui_automation_tests/step_defs/test_hmrc.py
```diff
@@ -38,3 +38,14 @@ def i_click_on_hmrc_set_end_user(driver):
 def i_click_on_hmrc_explain_your_reasoning(driver):
     app = ApplicationOverviewPage(driver)
     app.click_hmrc_explain_your_reasoning()
+
+
+@when("I wait for document to upload")
+def wait_for_document(driver):
+    document_is_found = False
+    while not document_is_found:
+        if 'Processing' in driver.find_element_by_id('document').text:
```
a-bek: Worth setting a `sleep(0.1)` in the while loop?

ui_automation_tests/step_defs/test_hmrc.py
phlptt: The refresh needs to be there for the text to change

ui_automation_tests/step_defs/test_hmrc.py
phlptt: ok

ui_automation_tests/step_defs/test_hmrc.py
andytaylorkainos: Yeah i know, but instead of refreshing the page potentially 50 times. We could use a `sleep` call to wait X seconds before refreshing once

PR #251, opened 2019-12-11
Replacing goods with products on select location page

anthony-lord:
Replacing 'goods' with 'products'


andytaylorkainos:
Covered by other PRs

PR #253, opened 2019-12-11
tool: HTML String Check

BenAndrew-Kainos:
String checker for finding HTML strings that are missing from lite_content.

Internal: https://github.com/uktrade/lite-internal-frontend/pull/274

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,5 +79,5 @@ jobs:
             pipenv run bandit -r .
 
       - run:
-          name: Run String checker
+          name: Run Content String Check
           command: pipenv run python templates/check_html_strings.py
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,3 +77,7 @@ jobs:
           command: |
             pipenv install bandit
             pipenv run bandit -r .
+
+      - run:
+          name: Run String checker
+          command: pipenv run python templates/check_html_strings.py
--- a/templates/check_html_strings.py
+++ b/templates/check_html_strings.py
@@ -27,7 +27,7 @@ def get_strings_package(base_dir):
     i.e. /Users/user/lite-internal-frontend = lite_internal_frontend
     :return: strings package name
     """
-    project_name = base_dir.split("/")[-1]
+    project_name = os.path.basename(base_dir)
     return project_name.replace("-", "_", 2)
 
 
@@ -58,9 +58,9 @@ def get_all_lcs_strings(templates_folder):
     strings = set()
 
     for root, _, files in os.walk(templates_folder):
-        for file in files:
-            if file.endswith(".html"):
-                with open(f"{root}/{file}") as f:
+        for file_name in files:
+            if file_name.endswith(".html"):
+                with open(f"{root}/{file_name}") as f:
                     lcs_strings = re.findall(LCS_PATTERN, f.read())
                     strings.update(lcs_strings)
 
--- a/templates/check_html_strings.py
+++ b/templates/check_html_strings.py
@@ -5,7 +5,7 @@ import re
 LCS_PATTERN = "{% lcs '(.*)' %}"
 
 
-def find_occurance(module, string):
+def check_string_for_occurance(module, string):
     """
     Check that the given string variable i.e. `CASES.Page.TITLE` is found within a given module (strings.py).
     Checks each section of the path is found inside the module defined in the last iteration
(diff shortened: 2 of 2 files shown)


templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
+import importlib
+import os
+import re
+
+LCS_REGEX = "{% lcs '(.*)' %}"
```
stephanieboyd: Let's call this `lcs_pattern`

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def find_occurance(module, string):
+    """
+    Check that the given string variable i.e. CASES.Page.TITLE is found within a given module (strings.py).
```
stephanieboyd: Backtick the variable name

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+    Checks each section of the path is found inside the module defined in the last iteration
+    :param module: Python module to search
+    :param string: Path to find (split into sections)
+    :return: bool of whether it was found
```
stephanieboyd: just write true/false here

.circleci/config.yml
```diff
@@ -77,3 +77,7 @@ jobs:
           command: |
             pipenv install bandit
             pipenv run bandit -r .
+
+      - run:
+          name: Run String checker
```
stephanieboyd: `Run content string check`

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+    strings_module = importlib.import_module(f"lite_content.{strings_package_name}.strings")
+
+    # Check all strings are found
+    lcs_strings = get_all_lcs_strings(templates_folder)
+    for string in lcs_strings:
+        if not find_occurance(strings_module, string.split(".")):
+            not_found.append(string)
+
+    if not_found:
+        raise Exception(f"The following strings couldn't be found: {not_found}")
```
stephanieboyd: I've left a lot of these comments on the one I did review - I don't think we should be raising an exception - just the script output is strings or no strings.

templates/check_html_strings.py
BenAndrew-Kainos: Done

templates/check_html_strings.py
BenAndrew-Kainos: Done

templates/check_html_strings.py
BenAndrew-Kainos: Done

.circleci/config.yml
BenAndrew-Kainos: done

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    for path_section in string:
+        module = module.__dict__.get(path_section)
+        if not module:
+            return False
+
+    return True
+
+
+def get_strings_package(base_dir):
```
stephanieboyd: there is a function for this, in os package.
```>>> os.path.basename("/Users/user/lite-internal-frontend")
'lite-internal-frontend'````

File: templates/check_html_strings.py [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+def get_all_lcs_strings(templates_folder):
+    """
+    Get all LCS strings from the HTML files in the given folder
+    :param templates_folder: string path for the HTML folders
+    :return: set of found LCS strings
+    """
+    strings = set()
+
+    for root, _, files in os.walk(templates_folder):
+        for file in files:
```
stephanieboyd: Can you change everywhere you've used the file keyword? Call it `file_name`

templates/check_html_strings.py
BenAndrew-Kainos: Done

templates/check_html_strings.py
BenAndrew-Kainos: Done

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+    # Check all strings are found
+    lcs_strings = get_all_lcs_strings(templates_folder)
+    for string in lcs_strings:
+        if not find_occurance(strings_module, string.split(".")):
+            not_found.append(string)
+
+    if not_found:
+        raise Exception(f"The following strings couldn't be found: {not_found}")
+    else:
+        print("All strings found")
```
stephanieboyd: "No unused strings found"?

templates/check_html_strings.py
```diff
@@ -0,0 +1,89 @@
...
+    templates_folder = f"{base_dir}/templates"
+
+    # Load strings package
+    strings_package_name = get_strings_package(base_dir)
+    strings_module = importlib.import_module(f"lite_content.{strings_package_name}.strings")
+
+    # Check all strings are found
+    lcs_strings = get_all_lcs_strings(templates_folder)
+    for string in lcs_strings:
+        if not find_occurance(strings_module, string.split(".")):
```
stephanieboyd: Can we call this `check_string_occurance`. Reads better in this condition, reads better that you're checking something

BenAndrew-Kainos: Done

templates/check_html_strings.py
BenAndrew-Kainos: Done

PR #257, opened 2019-12-11
Content review of goods locations

anthony-lord:
Replaced 'goods' with 'products'

### DIFF
--- a/applications/forms/told_by_an_official.py
+++ b/applications/forms/told_by_an_official.py
@@ -4,7 +4,7 @@ from lite_forms.components import Form, RadioButtons, Option, TextInput
 
 def told_by_an_official_form(application_id):
     return Form(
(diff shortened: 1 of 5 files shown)


applications/forms/told_by_an_official.py
```diff
@@ -16,9 +16,9 @@ def told_by_an_official_form(application_id):
                 classes=["govuk-radios--inline"],
             ),
             TextInput(
-                title="What was the reference number if you were provided one?",
-                description="This is the reference found on the letter or "
-                "email to tell you to apply for an export licence.",
+                title="Reference number (optional)",
+                description="The reference number is on the letter or "
+                "email.",
```
efahy: You should probably combine both parts of the description string onto one line, which would be `description="The reference number is on the letter or email.",`

PR #273, opened 2019-12-16
feat: LT-1450 - Paginate and filter goods

Mark-Spiers-kainos:


goods/services.py
```diff
@@ -13,7 +13,7 @@ def get_goods(request, params=None):
     else:
         data = get(request, GOODS_URL)
 
-    return data.json().get("goods"), data.status_code
```
jfaracik: Could you change get_goods to take filter params as opposed to a dict? Makes it more explicit. eg

```
# Organisation
def get_organisations(request, page: int = 1, search_term=None, org_type=None):
    """
    Returns a list of organisations
    :param request: Standard HttpRequest object
    :param page: Returns n page of page results
    :param search_term: Filter by name
    :param org_type: Filter by org type - 'hmrc', 'commercial', 'individual', or an array of it
    """
    data = get(request, ORGANISATIONS_URL + convert_parameters_to_query_params(locals()))
    return data.json()
```

Also can you remove status code

Mark-Spiers-kainos: okay, have updated this to take the filter params as you asked. Going to leave the status_code in for now as I want to tech share a small quality of life piece of code first.

templates/goods/goods.html
```diff
@@ -1,6 +1,7 @@
 {% extends 'layouts/base.html' %}
 
 {% load svg humanize %}
```
jfaracik: Move the static inside of this load tag

Mark-Spiers-kainos: done

PR #274, opened 2019-12-17
feat: LT-1201: Exporter post-submit audits

a-bek:


auth/backends.py
```diff
@@ -6,25 +6,24 @@
 class AuthbrokerBackend:
```
Mark-Spiers-kainos: why is the pr is touching authentication?

a-bek: Just cleaned the function up when I saw it, should I revert?

Mark-Spiers-kainos: I was just confused reading through first time as to why. But looks good.

PR #285, opened 2019-12-30
test: Extra step to verify build success

phlptt:


ui_automation_tests/step_defs/test_menu.py
```diff
@@ -0,0 +1,19 @@
+import urllib.request
+
+from pytest_bdd import scenarios, then, when
+
+scenarios("../features/menu.feature", strict_gherkin=False)
+
+
+@then("I get a 200")
+def success(driver):
```
efahy: Success won't always be the returning of a 200 and existence of a log out link.  I would rename the function to be more specific

ui_automation_tests/step_defs/test_menu.py
```diff
@@ -0,0 +1,19 @@
...
+scenarios("../features/menu.feature", strict_gherkin=False)
+
+
+@then("I get a 200 and log out link is displayed")
+def log_out_link_displayed_and_success_status_code(driver):
+    conn = urllib.request.urlopen(driver.current_url)
+    assert conn.getcode() == 200
+    assert driver.find_element_by_id(
+        "link-sign-out"
+    ).is_displayed(), "Log out button is displayed. User may have the service unavailable screen."
```
Mark-Spiers-kainos: Would this text not only appear if the log out button is not displayed?

phlptt: my bad, fixed

PR #286, opened 2020-01-02
Speed up 02022020

phlptt:
Speeding up tests, some are commented out and will only be commented back in if they are needed (ie if there is a regression in this part) 17m 03s down to 13m 42s

### DIFF
--- a/ui_automation_tests/features/case_notes.feature
+++ b/ui_automation_tests/features/case_notes.feature
@@ -8,8 +8,7 @@ Feature: I want to add a note to an application and view notes
   Scenario: Add a new valid case note
     Given I go to exporter homepage and choose Test Org
     And I create a standard application via api
-    When I click on applications
-    And I click on application previously created
+    When I go to application previously created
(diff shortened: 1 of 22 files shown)


ui_automation_tests/features/case_notes.feature
```diff
@@ -18,25 +17,14 @@ Feature: I want to add a note to an application and view notes
...
-  Scenario: Unable to add case note
-    Given I go to exporter homepage and choose Test Org
-    And I create a standard application via api
-    When I click on applications
-    And my application has been withdrawn
-    And I click on application previously created
-    And I click the notes tab
-    Then the case note text area is not present
\ No newline at end of file
+#  @LT_1119_too_many @regression
```
markm5A8Z9F43913A: Would be good to add a link to a JIRA ticket that will fix this.

markm5A8Z9F43913A: Same for other instances of commenting out tests in the PR

PR #287, opened 2020-01-02
Move exporter json strings to python

efahy:


applications/views/documents.py
```diff
@@ -11,24 +11,27 @@
...
 from lite_content.lite_exporter_frontend import strings
 from lite_forms.generators import form_page, error_page
 
 
 def get_upload_page(path, draft_id):
     paths = document_switch(path)
 
+    # Note the use of get_const_string for getting content strings by path.  Under notmal circumstances,
+    # we refer to strings directly, since they are Python class properties, however in this case the paths are
+    # built dynamically, based on party type so this method is necessary
```
BenAndrew-Kainos: I don't really like this. Could we instead extend the `document_switch` function to also return the title, back_link & description based on the path

templates/users/users.html
```diff
@@ -6,12 +6,6 @@
 
 {% block body %}
 
-	{% if userDeactivated == "true" %}
-		<div class="information information-animated">
-			{% get_string 'users.deleted' %}
```
BenAndrew-Kainos: Why was this removed? (just curious)

efahy: Its quite interesting actually: Your missing string finder tool reported `Users.DELETED` as missing when I ran my scripts and converted the above.  And when I investigated, it turned out the users.deleted string never existed, even when the above code was added.  So I checked with Jan and he said to just ditch it

applications/views/documents.py
efahy: `document_switch` extended as suggested 👍

applications/helpers/reverse_documents.py
```diff
@@ -28,53 +29,72 @@ def document_switch(path):
...
             "download": get_end_user_document,
             "delete": delete_end_user_document,
             "homepage": "applications:end_user",
-            "strings": "end_user.documents",
             "has_description": False,
+            "attach_doc_description_field_string": strings.EndUser.Documents.AttachDocuments.DESCRIPTION_FIELD_TITLE,
+            "attach_doc_title_string": strings.EndUser.Documents.AttachDocuments.TITLE,
+            "attach_doc_return_later_string": strings.EndUser.Documents.SAVE_END_USER,
+            "delete_conf_back_link_string": strings.EndUser.Documents.AttachDocuments.BACK_TO_APPLICATION_OVERVIEW,
+            # .AttachDocuments.DESCRIPTION_FIELD_TITLE
```
BenAndrew-Kainos: Please remove comment

templates/users/users.html
BenAndrew-Kainos: Awesome :D

PR #297, opened 2020-01-08
feat: LT-1232 - case officer assignment

Mark-Spiers-kainos:


templates/goods/good.html
```diff
@@ -51,6 +51,18 @@ <h3 class="govuk-heading-m">Your Query</h3>
 							{{ good.query.id|reference_code }}
 						</dd>
 					</div>
+                    <div class="govuk-summary-list__row">
```
efahy: Your lines are indented with spaces, the ones above and below are indented with tabs, which is why they look different here I guess.  Andrew tells me we've standardised on tabs, so can you change to that please?

PR #305, opened 2020-01-15
feat: LT-1200 - Reference code

jfaracik:
* We now use reference codes everywhere where it makes sense to do so
* Added copy button to reference code on Application page
* Application submission success page has been moved to its own view - following the https://en.wikipedia.org/wiki/Post/Redirect/Get pattern
* Removed reference_code template tag


templates/end-users/end-users.html
```diff
@@ -55,7 +55,7 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
 							{% endif %}
```
BenAndrew-Kainos: There are strings in this file that need to be moved to content: New end user advisory, Name, Type, Address, Website, Reference, Copy, There are no end user advisories., Information

templates/goods/good.html
```diff
@@ -65,7 +65,7 @@ <h3 class="govuk-heading-m">Your query</h3>
 							{% lcs 'goods.GoodPage.Query.REFERENCE' %}
```
BenAndrew-Kainos: There are strings in this file that need to be moved to content: If you're unsure the product is controlled or not, you can, raise a control list classification (CLC) query, Download, Delete

templates/applications/application.html
```diff
@@ -25,54 +25,69 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
...
+					<button class="lite-button--link lite-js-only" href="#" id="button-copy-reference-code">
+						{% lcs "applications.ApplicationSummaryPage.COPY_REFERENCE_CODE" %}
+					</button>
+				</dd>
+			</div>
+			<div class="govuk-summary-list__row">
+				<dt class="govuk-summary-list__key">
+					{% lcs 'applications.ApplicationSummaryPage.TYPE' %}
+				</dt>
+				<dd class="govuk-summary-list__value">
```
stephanieboyd: ID on this?

templates/applications/application.html
```diff
@@ -25,54 +25,69 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
...
+					{{ application.application_type.value }}
+				</dd>
+			</div>
+			{% if application.application_type.key != 'hmrc_query' %}
+			<div class="govuk-summary-list__row">
+				<dt class="govuk-summary-list__key">
+					{% lcs 'applications.ApplicationSummaryPage.EXPORT_TYPE' %}
+				</dt>
+				<dd class="govuk-summary-list__value">
```
stephanieboyd: ID on this?

templates/applications/application.html
```diff
@@ -25,54 +25,69 @@ <h1 class="govuk-heading-l">{{ title }}</h1>
...
+					{% else %}
+						{% lcs 'applications.ApplicationSummaryPage.NO_ASSIGNED_CASE_OFFICER' %}
+					{% endif %}
+				</dd>
+			</div>
+			<div class="govuk-summary-list__row">
+				<dt class="govuk-summary-list__key">
+					{% lcs 'applications.ApplicationSummaryPage.SUBMITTED_AT' %}
+				</dt>
+				<dd class="govuk-summary-list__value">
```
stephanieboyd: ID on this?

applications/urls.py
```diff
@@ -19,6 +19,7 @@
     path("", common.ApplicationsList.as_view(), name="applications"),
     path("<uuid:pk>/delete/", common.DeleteApplication.as_view(), name="delete"),
     path("<uuid:pk>/task-list/", common.ApplicationTaskList.as_view(), name="task_list"),
+    path("<uuid:pk>/completion/", common.ApplicationSubmitSuccessPage.as_view(), name="success_page"),
```
stephanieboyd: Hm, completion? I reckon the url should be more inline with "application success"

templates/applications/application.html
jfaracik: What benefit would that provide?

applications/views/common.py
```diff
@@ -124,7 +124,9 @@ def post(self, request, **kwargs):
         if status_code != HTTPStatus.OK:
             return get_application_task_list(request, application, errors=data.get("errors"))
 
-        return application_success_page(request, application_id)
+        # Redirect to the success page to prevent the user going back after the post
+        # Follows this pattern: https://en.wikipedia.org/wiki/Post/Redirect/Get
```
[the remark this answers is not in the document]
jfaracik: Why not?

core/helpers.py
```diff
@@ -78,7 +78,7 @@ def convert_parameters_to_query_params(dictionary: dict):
     if "request" in dictionary:
         del dictionary["request"]
 
-    return "?" + convert_dict_to_query_params({key: value for key, value in dictionary.items() if value})
+    return "?" + convert_dict_to_query_params({key: value for key, value in dictionary.items() if value is not None})
```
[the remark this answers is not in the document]
jfaracik: Drafts endpoint was failing as a result of this. submitted=false qualified and thus was being sent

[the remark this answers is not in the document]
jfaracik: It is.

templates/goods/good.html
jfaracik: I'm not doing "If you're unsure the product is controlled or no..." as the link can't be in content. Until we have a better solution for that sort of thing I'm leaving it.

core/helpers.py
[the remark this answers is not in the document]
BenAndrew-Kainos: Ok

PR #306, opened 2020-01-15
bug: LT-1889 CLC Query Generated Documents

BenAndrew-Kainos:


templates/goods/good.html
```diff
@@ -220,6 +220,9 @@ <h3 class="govuk-heading-m">{% lcs 'goods.GoodsList.Documents.TITLE' %}</h3>
 			                        {{ good.query.exporter_user_notification_count.ecjuquery }}
 			                    </div>
 						    {% endif %}
+			            </a>
+						<a href="{% url 'goods:good_detail' good.id 'ecju-generated-documents' %}" class="lite-tabs__tab {% if type == 'ecju-generated-documents' %}lite-tabs__tab--selected{% endif %}" id="link-ecju-generated-documents">
+			                {% lcs 'goods.GoodPage.Tabs.GENERATED_DOCUMENTS' %}
```
efahy: Can you standardise on tab chars for indenting please?

templates/goods/tabs/ecju-generated-documents.html
```diff
@@ -0,0 +1,13 @@
+{% if generated_documents %}
```
efahy: If there are no generated docs, would be good to put a line of text to indicate that, rather than having an empty tab

templates/goods/tabs/ecju-generated-documents.html
```diff
@@ -0,0 +1,13 @@
...
+    {% for generated_document in generated_documents %}
+        <div class="app-ecju-query">
+            <div class="app-ecju-query__item">
+                <p class="app-ecju-query__heading">{{ generated_document.name }}</p>
+                <p class="app-ecju-query__date">
+                    <span class="govuk-visually-hidden">{% lcs 'goods.GoodPage.ECJUDocuments.CREATED_AT' %}</span>{{ generated_document.created_at|str_date }}
```
efahy: Currently, the doc name is not a link that allows exporters to view the doc, its just the text of the doc name which seems pointless to me.  Can you check with the product team to see if this is how its meant to be please?

templates/goods/good.html
BenAndrew-Kainos: Done

templates/goods/tabs/ecju-generated-documents.html
BenAndrew-Kainos: Done

templates/goods/tabs/ecju-generated-documents.html
BenAndrew-Kainos: Done

core/services.py
```diff
@@ -164,3 +168,11 @@ def get_control_list_entries(request, convert_to_options=False):
 def get_control_list_entry(request, rating):
     data = get(request, CONTROL_LIST_ENTRIES_URL + rating)
     return data.json().get("control_list_entry")
+
+
+def get_document_download(request, url):
+    response = get(request, url)
+    if response.status_code == 200:
```
stephanieboyd: Use HTTPStatus please

goods/services.py
```diff
@@ -57,12 +60,21 @@ def raise_clc_query(request, json):
...
+    return data.json(), data.status_code
+
+
 # Documents
 def get_good_document(request, pk, doc_pk):
     data = get(request, GOODS_URL + pk + DOCUMENTS_URL + doc_pk)
     return data.json().get("document") if data.status_code == HTTPStatus.OK else None
 
 
+def get_case_document_download(request, file_pk, case_pk):
```
stephanieboyd: Same minor comment as API. file -> document would be better

core/services.py
BenAndrew-Kainos: Done

goods/services.py
BenAndrew-Kainos: Done

core/services.py
```diff
@@ -164,3 +168,11 @@ def get_control_list_entries(request, convert_to_options=False):
 def get_control_list_entry(request, rating):
     data = get(request, CONTROL_LIST_ENTRIES_URL + rating)
     return data.json().get("control_list_entry")
+
+
+def get_document_download(request, url):
```
efahy: Maybe `get_document_download_stream`?

core/services.py
```diff
@@ -164,3 +168,11 @@ def get_control_list_entries(request, convert_to_options=False):
...
+    response = get(request, url)
+    if response.status_code == HTTPStatus.OK:
+        return StreamingHttpResponse(response, content_type=response.headers._store["content-type"][1])
+    else:
+        return error_page(request, Document.ACCESS_DENIED)
```
efahy: I think its a bit dangerous to always assume that its always access denied if the response isn't ok.  Perhaps check specifically for access denied response in an else if and give details of the actual error in an else?

goods/services.py
```diff
@@ -87,3 +97,10 @@ def get_document_missing_reasons(request):
 def post_good_document_sensitivity(request, pk, json):
     data = post(request, GOODS_URL + str(pk) + DOCUMENT_SENSITIVITY_URL, json)
     return data.json(), data.status_code
+
+
+# CLC Query
```
efahy: Is this ONLY CLC case documents?

BenAndrew-Kainos: Currently yes but hopefully will expand over time

core/services.py
BenAndrew-Kainos: Done

core/services.py
BenAndrew-Kainos: I've expanded this to show this error if 401 (Unauthorised) and show a different more generic error otherwise

goods/services.py
efahy: I don't think this comment is adding any value.  If anything, its confusing things so I'd either remove it or explain the situation more fully

BenAndrew-Kainos: Removed

PR #310, opened 2020-01-22
test: Add CLC query test and some further small changes

phlptt:


ui_automation_tests/pages/goods_list.py
```diff
@@ -21,6 +21,14 @@ def assert_clc_goods_are_displayed_of_good_name(self, driver, description, part_
         assert control_code in goods_row
         assert "Control List Classification Query" in goods_row
 
+    def assert_clc_goods_is_displayed_on_good_page(self, driver, description, part_number, control_code):
+        goods_row = Shared(driver).get_text_of_gov_table()
+
+        assert description in goods_row
+        assert part_number in goods_row
+        assert control_code in goods_row
+        assert "Control List Classification Query" in goods_row
```
stephanieboyd: Why is this the only assert that doesn't get it's value passed to the function?

PR #317, opened 2020-01-27
Content changes 27 01 20

anthony-lord:
Changed 'Completed' to 'Saved'
Changed 'In progress' to 'Add document'

### DIFF
--- a/templates/applications/standard-application-edit.html
+++ b/templates/applications/standard-application-edit.html
@@ -103,7 +103,7 @@
 						</a>
 						{% if sites or external_locations %}
 							<div class="lite-tag lite-tag--blue">
-								Completed
+								Saved
 							</div>
 						{% else %}
 							<div class="lite-tag lite-tag--hollow">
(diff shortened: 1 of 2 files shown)


templates/applications/standard-application-edit.html
```diff
@@ -251,11 +251,11 @@ <h2 class="lite-task-list__section">
                         {% endif %}
 						{% if application.consignee and consignee_document %}
 							<div class="lite-tag lite-tag--blue">
-								Completed
+								Saved
 							</div>
 				        {% elif application.consignee %}
 				            <div class="lite-tag lite-tag--orange">
-								In progress
+								Add document
```
jfaracik: Task list items are supposed to show the status of each item, eg "Done", "In progress"

I could imagine "Add document" confusing users into thinking that the tag is a link.

anthony-lord: User research feedback shows that 'Done' or 'Completed' is confusing to our users as they have only added 1 of 5 products for example, but the task list is telling them that section is done but it isn't. Also users aren't sure if their application is saved when they look at the task list as there's no 'Save' button. So we're trying to solve both those issues by changing the label to 'Saved'.
Daniel was concerned that 'In progress' might imply that that section hasn't been saved when next to a section with the 'Saved' label, so we're gonna try 'Add document'. Agree that 'Add' could suggest it's a button, but other options for that content are too long. As production is our prototype, we'll see what happens in research and revert back to 'In progress' if users think it's a button

jfaracik: That's understandable, when the application page refactor comes in I'm certain we'll add a button to mark each section as "Done" which will hopefully solve that issue.

PR #318, opened 2020-01-29
feat: Lt 1874 remove duplicate people from previously referenced list

stephanieboyd:


templates/applications/parties/preexisting.html
```diff
@@ -49,6 +49,11 @@ <h1 class="govuk-heading-l">{% lcs 'applications.CopyExistingPartyPage.HEADING'
                                     <input type="hidden" name="website" value="{{ party.website }}">
                                     <input type="hidden" name="address" value="{{ party.address }}">
                                     <input type="hidden" name="country" value="{{ party.country.id }}">
+									{% if party.copy_of %}
+										<input type="hidden" name="copy_of" value="{{ party.copy_of }}">
```
andytaylorkainos: you'll probably need a label for these inputs or axe might complain

applications/services.py
```diff
@@ -153,6 +153,14 @@ def delete_ultimate_end_user(request, pk, obj_pk):
     return data.status_code
 
 
+def validate_ultimate_end_user(request, pk, json):
+    json = json.copy()
+    json["validate_only"] = True
+
+    data = post(request, APPLICATIONS_URL + str(pk) + "/ultimate-end-users/", json)
```
BenAndrew-Kainos: Please move "/ultimate-end-users/" to constant

applications/views/parties/third_parties.py
```diff
@@ -31,13 +31,18 @@ def __init__(self):
...
             name="third_party",
-            action=post_third_party,
             form=third_party_forms,
             back_url="applications:add_third_party",
             strings=ThirdPartyForm,
             multiple_allowed=True,
         )
 
+    def on_submission(self, request, **kwargs):
+        if int(self.request.POST.get("form_pk")) == len(self.forms.forms) - 1:
```
BenAndrew-Kainos: Please move this logic to the parent class and pass a validate and post function to the parent (so that this logic is not duplicated elsewhere)

applications/services.py
stephanieboyd: Moved, I also moved a bunch more in this file that needed it

applications/views/parties/third_parties.py
stephanieboyd: It only applies to 2 types of parties though?

BenAndrew-Kainos: Currently yes although this bug is technically present on end user & consignee (you just don't realise because it overwrites instead of adds). To keep in line with the current structure I'd like all party logic in the parent class

stephanieboyd: I've added validate only for consignee and end users also

PR #319, opened 2020-01-30
feat: LT-1528 - respond to pv grading

Mark-Spiers-kainos:


templates/goods/good.html
```diff
@@ -50,13 +50,12 @@ <h1 class="govuk-heading-l">
 			</div>
 
 			<!-- Notify the user if their good has a special status -->
-			{% if good.status.key == 'clc_query' %}
+			{% if good.status.key == 'query' %}
```
Mark-Spiers-kainos: This fixes the banner being missing bug to tell users they are currently in a query.

PR #321, opened 2020-01-30
feat: LT-1084 - Record details of when a customs enquiry is being performed

jfaracik:
* Removed the "Is your good controlled?" question for HMRC apps
* Updated to the latest shared assets and forms
* Simplified HMRC E2E
* Fixed goods locations appearing as already filled in
* Fixed countries matrix task not showing
* Fixed third parties showing as completed on standard applications

this fucking cat always gets me to get in a good mood - Imgur


applications/helpers/validators.py
```diff
@@ -25,8 +25,18 @@ def validate_delete_draft(request, pk, json):
     return {"errors": {"choice": [strings.applications.DeleteApplicationPage.DELETE_ERROR]}}, HTTPStatus.BAD_REQUEST
 
 
+def validate_goods_location_choice(_request, _pk, json):
+    choice = json.get("choice")
+
+    if choice:
+        put_application(_request, _pk, {"is_goods_departed": choice == "departed"})
```
andytaylorkainos: can you add an enum for these choices and update where appropriate please

jfaracik: Done

apply_for_a_licence/forms.py
```diff
@@ -37,7 +36,6 @@ def opening_question():
         questions=[
             RadioButtons(
                 name="licence_type",
-                title=InitialApplicationQuestionsForms.OpeningQuestion.LABEL,
```
BenAndrew-Kainos: Removing titles breaks accessibility

jfaracik: I'll be fixing single component pages in a future forms update.

BenAndrew-Kainos: Could we leave this until then :)

applications/helpers/validators.py
```diff
@@ -25,8 +26,18 @@ def validate_delete_draft(request, pk, json):
     return {"errors": {"choice": [strings.applications.DeleteApplicationPage.DELETE_ERROR]}}, HTTPStatus.BAD_REQUEST
 
 
+def validate_goods_location_choice(_request, _pk, json):
```
efahy: This function `put`s the application, it doesn't just validate it - please rename

jfaracik: What would you have me rename it to?

efahy: How about `update_goods_location_choice`?  But I now realise its in `validators.py`, which doesn't seem right either

efahy: I guess you need a method in `applications/services.py`?

PR #325, opened 2020-02-03
chore: Clean up lite-tests-common

andytaylorkainos:


ui_automation_tests/fixtures/add_goods_query.py
```diff
@@ -0,0 +1,10 @@
+from pytest import fixture
+
+from shared.tools.utils import get_lite_client
+
+
+@fixture(scope="function")
+def add_goods_query(context, api_client_config):
+    lite_client = get_lite_client(context, api_client_config=api_client_config)
+    lite_client.goods_queries.add_clc_good(lite_client.goods, lite_client.ecju_queries)
```
efahy: For completeness, would it be good to add a PV grading element too?

andytaylorkainos: renamed function to reflect to code

PR #326, opened 2020-02-04
feat: LT-1966 - Case notes styling

jfaracik:
_What's changed_
* Updated styling for case note to be "inline" with GDS design system
* Added notice to the activity tab when there's no activity
* Updated E2E

dPOUXHx


templates/applications/tabs/activity.html
```diff
@@ -17,4 +17,6 @@
 			{% endfor %}
 		</div>
 	</div>
+{% else %}
+	{% include "includes/notice.html" with text="REPLACE ME" %}
```
cstobart: good spot

PR #329, opened 2020-02-05
feat: LT-979 surrender a licence

BenAndrew-Kainos:


applications/helpers/validators.py
```diff
@@ -17,6 +17,18 @@ def validate_withdraw_application(request, pk, json):
     )
 
 
+def validate_surrender_application(request, pk, json):
```
efahy: As I commented on a recent PR from Jan, this function name is misleading because it doesn't make clear that its not just validation that it does - it also actually changes the case's status.  Jan was reluctant to move his function despite the fact that I saw it as more of a service method, so we ended up renaming it.  In this instance, what about changing the name to `validate_surrender_application_and_update_case_status`? Too long? :)

applications/helpers/validators.py
```diff
@@ -17,6 +17,18 @@ def validate_withdraw_application(request, pk, json):
...
+    if json.get("choice"):
+        if json.get("choice") == "yes":
```
efahy: Minor one - should you store the result of the `json.get` in a local var to avoid having to do it twice?

templates/applications/application.html
```diff
@@ -17,14 +17,19 @@ <h1 class="govuk-heading-l">
...
+				    <a href="{% url 'applications:withdraw' application.id %}" id="button-withdraw-application" class="govuk-button govuk-button--secondary">
+					    {% lcs 'applications.ApplicationSummaryPage.Buttons.WITHDRAW_ACCESS_BUTTON' %}
+				    </a>
+                {% endif %}
+                {%  if not status_is_read_only %}
+				    <a href="{% url 'applications:edit_type' application.id %}" id="button-edit-application" class="govuk-button govuk-button--secondary">
+					    {% lcs 'applications.ApplicationSummaryPage.Buttons.EDIT_APPLICATION_BUTTON' %}
+				    </a>
+                {% endif %}
+				{% if application.licence_duration %}
```
efahy: Might be worth a comment to say that really the condition is that there are approve or proviso decisions but that you're using license_duration as a shortcut - it might prevent mishaps in future?  I assume that's the thinking anyway

applications/views/common.py
```diff
@@ -296,6 +300,25 @@ def init(self, request, **kwargs):
         self.success_url = reverse_lazy("applications:application", kwargs={"pk": self.object_pk})
 
 
+class SurrenderApplication(SingleFormView):
+    def init(self, request, **kwargs):
+        self.object_pk = kwargs["pk"]
+        application = get_application(request, self.object_pk)
+        self.form = confirm_form(
+            title="Surrender?",
```
efahy: Lite content for all on screen text in here??

applications/helpers/validators.py
BenAndrew-Kainos: Done

applications/helpers/validators.py
BenAndrew-Kainos: Yep

templates/applications/application.html
BenAndrew-Kainos: Added the comment but we wouldn't consider it a shortcut

applications/views/common.py
BenAndrew-Kainos: Done

applications/views/common.py
```diff
@@ -296,6 +300,25 @@ def init(self, request, **kwargs):
         self.success_url = reverse_lazy("applications:application", kwargs={"pk": self.object_pk})
 
 
+class SurrenderApplication(SingleFormView):
+    def init(self, request, **kwargs):
+        self.object_pk = kwargs["pk"]
+        application = get_application(request, self.object_pk)
+        self.form = confirm_form(
```
Mark-Spiers-kainos: can this call a function in the forms.py that returns the form, to keep them all in one place?

BenAndrew-Kainos: done

PR #339, opened 2020-02-11
chore: Cci poc

mh-hmcts:


ui_automation_tests/pages/exporter_hub_page.py
```diff
@@ -40,8 +41,11 @@ def login(self, email, password):
         if "logout" in self.driver.current_url:
             self.driver.find_element_by_xpath("//a[text()[contains(.,'Log In')]]").click()
 
-        if self.driver.find_elements_by_link_text("Accept all cookies"):
-            self.driver.find_element_by_link_text("Accept all cookies").click()
+        try:
```
phlptt: @Mark-Spiers-kainos reckons that the way on the left (note that it got changed to find_elements_by_link_text already) is quicker - that the try and catch as it has a 10 second timeout. I'm happy with whatever way people want to do this, should just choose one way for consistency.

mh-hmcts: agreed

PR #340, opened 2020-02-12
feat: LT-1229 - Anti-piracy

jfaracik:
* Cleaned up duplicate forms for editing applications
* Added form for goods categories
* Moved some strings from strings.py to generic.py
* Added `disabled` option for task list items
* Added multiple template tags: `join_list`, `join_key_value_list` and `equals`
* EXTRA: Added icon to readme


applications/views/misc.py
```diff
@@ -0,0 +1,38 @@
+from django.urls import reverse_lazy
```
a-bek: What do you think of `/views/edit.py`

ui_automation_tests/pages/apply_for_a_licence_page.py
```diff
@@ -39,6 +39,11 @@ def click_export_licence(self, export_type):
         elif export_type == "open":
             return self.driver.find_element_by_css_selector(self.BUTTON_OPEN_LICENCE).click()
 
+    def select_goods_categories(self):
+        checkboxes = self.driver.find_elements_by_name("goods_categories[]")
```
phlptt: This doesn't confirm that those checkboxes are there if there arent any it will just continue. Could you change it to self.driver.find_elements_by_name("goods_categories[]")[0].click self.driver.find_elements_by_name("goods_categories[]")[1].click
self.driver.find_elements_by_name("goods_categories[]")[2].click
self.driver.find_elements_by_name("goods_categories[]")[3].click

phlptt: or add an assert that there are 4

jfaracik: I'll add an assert, good spot

applications/views/misc.py
jfaracik: Done

PR #348, opened 2020-02-17
feat: LT-1249 - Add note to a draft application

jfaracik:
* Added page for draft application notes
* Added section to ask list for all application types but HMRC Query
* Renamed `post_application_case_notes` to `post_case_notes` for consistency
* Made case note errors inline 
* Fixed list number styling
* Added E2E test, cleaned up some others


applications/views/common.py
```diff
@@ -300,6 +291,37 @@ def init(self, request, **kwargs):
         self.success_url = reverse_lazy("applications:application", kwargs={"pk": self.object_pk})
 
 
+class Notes(TemplateView):
+    application_id = None
+    application = None
+
+    def dispatch(self, request, *args, **kwargs):
```
a-bek: Can we move this into `get` only as `post` does not need to load application.

jfaracik: Done!

ui_automation_tests/pages/submitted_applications_page.py
```diff
@@ -18,19 +18,29 @@ def enter_case_note(self, text):
...
         return self.driver.find_element_by_id(self.INPUT_CASE_NOTE_ID).text
 
-    def click_post_note_btn(self):
-        time.sleep(0.5)
+    def click_post_note_button(self):
         self.driver.find_element_by_id(self.BUTTON_POST_NOTE_ID).click()
 
-    def click_cancel_btn(self):
-        time.sleep(0.5)
+    def click_cancel_button(self):
```
phlptt: dont think this is actualyl used anymore

PR #351, opened 2020-02-24
feat: LT-1594 - Application and party clearance levels

a-bek:
- F380 application requires clearance level
- F380 parties require clearance level


applications/views/clearance.py
```diff
@@ -0,0 +1,27 @@
...
+from applications.services import put_application, get_application
+from core.services import get_pv_gradings
+from lite_forms.components import Form, RadioButtons
+from lite_forms.views import SingleFormView
+
+
+def clearance_level_form(application_id, options):
+    return Form(
+        title="Select which level of clearance.",
+        description="Clearance is important",
```
iichr: should these be retrieved from content?

applications/helpers/check_your_answers.py
```diff
@@ -184,71 +168,43 @@ def _convert_countries(countries):
     return [{"Name": country["name"]} for country in countries]
 
 
-def convert_party(party, application_id, editable):
+def convert_party(party, application, editable):
```
iichr: 👍 nice refactoring, logic is a lot clearer now

PR #354, opened 2020-02-25
bug: LT-2155 - HMRC query content strings missing

iichr:
- task list now appears properly
https://uktrade.atlassian.net/browse/LT-2155


iichr:
Closing as now addressed

PR #359, opened 2020-02-28
Tidy up 2802

phlptt:
Tidy up of some tests:
* mostly removing unused functions
* moving stuff from conftest thats only used in one feature file (not these changes are out of scope as they are just getting moved)
* Renamed some steps

Green build https://jenkins.ci.uktrade.io/view/LITE/job/lite-e2e-exporter-frontend/975/console


ui_automation_tests/conftest.py
```diff
@@ -630,7 +429,7 @@ def the_good_is_added_to_the_application(driver, context):  # noqa
     functions.click_back_link(driver)
 
 
-@then("wait for download link")  # noqa
+@when("I wait for download link")  # noqa
```
BenAndrew-Kainos: Why is this now a when if it does an assertion. Shouldn't this be "Then the download link is present"

ui_automation_tests/step_defs/test_standard_application.py
```diff
@@ -184,3 +165,82 @@ def filter_for_party(driver, context):
...
+    functions.click_submit(driver)
+    enter_export_licence(driver, "yes", "123456", context)
+
+
+@then("I see the application overview")  # noqa
+def i_see_the_application_overview(driver, context):  # noqa
+    element = TaskListPage(driver).get_text_of_lite_task_list_items()
+    assert context.app_name in element
+
+    app_id = driver.current_url[-36:]
```
BenAndrew-Kainos: Wth is this -36

ui_automation_tests/conftest.py
phlptt: ok it should probably be renamed to then download link is present

ui_automation_tests/step_defs/test_standard_application.py
phlptt: lol i just moved this from conftest to here as it was the only place its used. It's how its getting the ID - get thej url and get the last 36 characters ie the application ID

PR #369, opened 2020-03-06
fix: LT-2131 - display document name instead of download on case page

cstobart:


applications/helpers/check_your_answers.py
```diff
@@ -282,15 +282,15 @@ def _convert_document(party, document_type, application_id, editable):
     if editable:
         return convert_to_link(
             f"/applications/{application_id}/{document_type}/{party['id']}/document/download",
-            "Download",
+            Parties.Documents.DOWNLOAD + " " + document["name"],
```
a-bek: ```f"{Parties.Documents.DOWNLOAD} {document['name']}"```

jfaracik: Mmm f strings

PR #373, opened 2020-03-10
feat: LT-1233 - WMD end use concerns section

andytaylorkainos:


applications/views/end_use_details.py
```diff
@@ -0,0 +1,46 @@
...
+class EndUseDetails(SummaryListFormView):
+    def init(self, request, **kwargs):
+        self.object_pk = kwargs["pk"]
+        application = get_application(request, self.object_pk)
+        self.forms = end_use_details_form(application, request)
+        self.action = put_application
+        self.data = self._parse_end_use_details(application)
+        self.success_url = reverse_lazy("applications:task_list", kwargs={"pk": self.object_pk})
+        self.summary_list_title = strings.EndUseDetailsSummaryList.TITLE
+        self.summary_list_notice_title = ""
```
stephanieboyd: Do these get used or set to anything else anywhere? Did a search and it doesn't come up anywhere

iichr: I _believe_ lite_forms may have default values for some of them, so we override

PR #385, opened 2020-03-17
test: e2e test for registering an organisation and seeding user per test run

phlptt:


ui_automation_tests/conftest.py
```diff
@@ -68,6 +68,13 @@
 
 
 def pytest_addoption(parser):
+    settings.configure()
```
kowalcj0: @phlptt this might be causing that `DJANGO_SETTINGS_MODULE` error

Pipfile
```diff
@@ -16,6 +16,8 @@ pytest-html = "~=2.0"
 pytest-rerunfailures = "~=8.0"
 pytest-xdist = "~=1.31"
 selenium = "~=3.141"
+directory-sso-api-client = "*"
```
Mark-Spiers-kainos: I thought we were setting the pipfile to exact versions

BenAndrew-Kainos: Yeah please do that

pipeline/Jenkinsfile
```diff
@@ -15,7 +15,7 @@ spec:
     role: worker
   containers:
   - name: lite-selenium-test
-    image: quay.io/uktrade/lite-selenium-base
+    image: quay.io/uktrade/lite-selenium-base:morelibs
```
BenAndrew-Kainos, line 18: Point this back to master when you push

Pipfile
phlptt: done thanks

ui_automation_tests/conftest.py
```diff
@@ -1,6 +1,6 @@
 import datetime
 import os
-
+from django.conf import settings
```
jfaracik, line 3: Does this still refer to our frontends settings file?

PR #386, opened 2020-03-20
feat: LT-1060 - Non-UK organisations and sites

jfaracik:
**Story changes**
* Users can now register non-UK organisations right from the start page
* Organisations can now have sites in non-UK countries
* It's now possible to assign users to a site right from the creation flow

**Performance and misc improvements**
* Moved `new_site_form` over to `MultiFormView` from `TemplateView`
* Cleaned up organisation views


core/forms.py
```diff
@@ -74,39 +77,81 @@ def register_triage():
...
+    from core.views import RegisterAnOrganisationTriage
+
+    is_in_uk = location == RegisterAnOrganisationTriage.Locations.UNITED_KINGDOM
+
     return Form(
         title=conditional(
             not is_individual,
-            RegisterAnOrganisation.Headquarters.TITLE,
-            RegisterAnOrganisation.Headquarters.TITLE_INDIVIDUAL,
+            conditional(
```
Mark-Spiers-kainos: I don't like reading this

templates/layouts/base.html
```diff
@@ -104,9 +105,13 @@
 					</span>
 				</p>
 			</div>
-			{% block back_link %}
-				<a href="#" id="back-link" class="govuk-back-link">Back</a>
-			{% endblock %}
+			{% if request.GET|get:'return_to' %}
+				<a href="{{ request.GET|get:'return_to' }}" id="back-link" class="govuk-back-link">Back</a>
```
Mark-Spiers-kainos: lite-content

ui_automation_tests/step_defs/test_sites.py
```diff
@@ -18,49 +22,46 @@ def click_new_site(driver):
...
+    functions.click_submit(driver)
 
 
-@when("I click the first view link")
-def click_first_view_link(driver):
-    site_list_overview_page = SitesListOverview(driver)
-    site_list_overview_page.click_on_the_view_button_at_first_position()
+@when("I assign all users")
+def assign_all_users(driver, context):
+    NewSite(driver).click_all_users()
```
phlptt: can you add an assert or something here as all this will do if there are no users is loop through nothing

ui_automation_tests/step_defs/test_sites.py
```diff
@@ -18,49 +22,46 @@ def click_new_site(driver):
...
+    functions.click_submit(driver)
 
 
-@when("I click the edit button")
-def click_edit_button(driver):
-    SitePage(driver).click_edit_button()
+@then("the site is created")
+@then("the site is updated")
+def site_is_created_updated(driver, context):
+    assert context.new_site_name in driver.title
```
phlptt: would rather this assert was checking something on the page to make sure the page wasn't falling over

PR #389, opened 2020-03-24
Lt 2077 route of goods

stephanieboyd:


applications/views/route_of_goods.py
```diff
@@ -0,0 +1,34 @@
...
+        self.success_url = reverse_lazy("applications:task_list", kwargs={"pk": self.object_pk})
+
+    def get_form_data(self, request):
+        application = get_application(request, self.object_pk)
+        data = {}
+        if application["is_shipped_waybill_or_lading"] is not None:
+            data["is_shipped_waybill_or_lading"] = application["is_shipped_waybill_or_lading"]
+
+        if application["non_waybill_or_lading_route_details"] is not None:
+            data["non_waybill_or_lading_route_details"] = application["non_waybill_or_lading_route_details"]
```
iichr: since this is a string, I believe just `if application["non_waybill_or_lading_route_details"]` would suffice, otherwise empty strings would also pass - `"" is not None` evaluates to True

ui_automation_tests/pages/route_of_goods_form_page.py
```diff
@@ -0,0 +1,23 @@
...
+    def click_on_yes_radiobutton(self):
+        self.driver.find_element_by_css_selector("[id$=-True]").click()
+
+    def click_on_no_radiobutton(self):
+        self.driver.find_element_by_css_selector("[id$=-False]").click()
+
+    def enter_non_waybill_or_lading_route_details(self, details):
+        details_field = self.driver.find_element_by_class_name(self.TEXT_AREA_CLASS)
+        details_field.clear()
+        details_field.send_keys(details)
```
iichr: Ooh does that work for your textarea? I had some problems with it and had to use execute_script instead

applications/helpers/task_list_sections.py
```diff
@@ -58,3 +58,9 @@ def get_end_use_details(application):
         if application.get(field) is None:
             return False
     return True
+
+
+def get_route_of_goods(application):
+    if application["is_shipped_waybill_or_lading"] is None:
```
andytaylorkainos: does this field exist on all application types?
if not, it might be better to use `.get()`

andytaylorkainos: also. what would happen if `application["is_shipped_waybill_or_lading"]` was False?

ui_automation_tests/pages/route_of_goods_form_page.py
andytaylorkainos: same for me 🤷‍♂

applications/helpers/task_list_sections.py
stephanieboyd: This function is only used under the standard/open sections of the task list so it shouldn't happen but I will make it more robust. If the section is false, it's still been answered so it's fine :)

ui_automation_tests/pages/route_of_goods_form_page.py
stephanieboyd: Yeah, it works. I've got a nice screenshot. 
image

applications/views/route_of_goods.py
stephanieboyd: changed

ui_automation_tests/conftest.py
```diff
@@ -259,6 +260,15 @@ def eu_compliant_limitations_end_use_details(driver, choice):  # noqa
     functions.click_submit(driver)
 
 
+@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading'))  # noqa
+def route_of_goods(driver, choice):  # noqa
+    route_of_goods = RouteOfGoodsFormPage(driver)
+    if choice == "No":
+        route_of_goods.answer_route_of_goods_question(True, "Not shipped air waybill.")
```
phlptt: are these booleans the wrong way round True is no and False is yes

stephanieboyd: The true/false is just used to conditionally provide a details field also - which happens for No. So all good!

PR #402, opened 2020-04-06
feat: LT-2270 - Optional document upload

iichr:


applications/views/documents.py
```diff
@@ -46,26 +53,35 @@ def get(self, request, **kwargs):
...
 
-        logging.info(self.request)
-        data, error = add_document_data(request)
-
-        if error:
-            return form_page(request, form, extra_data={"draft_id": draft_id}, errors={"documents": [error]})
+        if (
+            request.FILES
+            or ("/end-user" in request.path and _is_application_export_type_permanent(application))
+            or "additional-document" in request.path
```
andytaylorkainos, line 64: can you add a comment above to explain th conditon? :)

iichr: Done

applications/helpers/check_your_answers.py
```diff
@@ -413,3 +414,7 @@ def get_total_goods_value(goods: list):
 
 def _is_application_export_type_temporary(application):
     return application.get("export_type").get("key") == TEMPORARY
+
+
+def is_application_export_type_permanent(application):
+    return False if not application.get("export_type") else (application.get("export_type").get("key") == PERMANENT)
```
iichr, line 420: this is also called by MOD applications hence the need to check for an export type

applications/helpers/reverse_documents.py
```diff
@@ -44,6 +48,7 @@ def document_switch(path):
         }
     elif "additional-document" in path:
         return {
+            "optional": False,
```
iichr, line 51: Additional documents must be provided if the section is filled in

applications/views/parties/end_users.py
```diff
@@ -25,10 +26,10 @@ def get(self, request, **kwargs):
                     application=application,
                     editable=application["status"]["value"] == "draft",
                 ),
-                "highlight": ["Document"] if not application["end_user"]["document"] else {},
+                "highlight": ["Document"] if (is_permanent_app and not application["end_user"]["document"]) else {},
```
iichr, line 29: A document must be uploaded in end user only if the application is standard permanent

applications/forms/documents.py
```diff
@@ -5,11 +5,14 @@
 from lite_forms.generators import confirm_form
 
 
-def attach_document_form(application_id, strings, back_link):
+def attach_document_form(application_id, strings, back_link, optional):
```
jfaracik, line 8: Should be is_optional in the param :) `ngs, back_link, is_optional):`

applications/helpers/check_your_answers.py
jfaracik: Could you do `return application.get("export_type", {}).get("key") == PERMANENT`

iichr: Not possible I'm afraid ☹️ returns None if it's not found and then the subsequent `.get` fails as unknown attribute

applications/forms/documents.py
iichr: Sorted 💯

applications/helpers/check_your_answers.py
iichr: Could change to multi-line if you reckon it'd be clearer, seems too verbose though
```
export_type = application.get("export_type")
if export_type:
  return export_type.get.... == PERMANENT
return False

jfaracik: Nah I think it's fine the way it is 🚀

PR #409, opened 2020-04-15
feat: LT-1021 Licensed Quantity & Licence page

BenAndrew-Kainos:


core/builtins/custom_tags.py
```diff
@@ -60,6 +62,29 @@ def str_date(value):
     )
 
 
+@register.filter
+@stringfilter
+def str_date_only(value):
+    date_str = do_timezone(datetime.datetime.strptime(value, DATE_FORMAT), "Europe/London")
```
andytaylorkainos, line 68: can you make "Europe/London" a constant please

core/builtins/custom_tags.py
```diff
@@ -60,6 +62,29 @@ def str_date(value):
...
+    return date_str.strftime("%d %B %Y")
+
+
+@register.filter()
+def add_months(start_date, months):
+    start_date = datetime.datetime.strptime(start_date, "%Y-%m-%d")
```
andytaylorkainos, line 74: not going to use your DATE_FORMAT constant?

core/builtins/custom_tags.py
```diff
@@ -60,6 +62,29 @@ def str_date(value):
...
+    year = start_date.year
+    month = start_date.month
+
+    for _ in range(months):
```
andytaylorkainos, line 78: really surprised that python's `datetime` doesnt have an easy way of doing this 🤯

core/builtins/custom_tags.py
```diff
@@ -387,3 +412,8 @@ def abbreviate_string(string, length):
         return string
     else:
         return str(string)[:length] + "..."
+
+
+@register.filter()
+def get_usage_value(good):
+    return (good["usage"] / good["licenced_quantity"]) * float(good["licenced_value"])
```
andytaylorkainos, line 419: does this need to be rounded in any way?

licences/views.py
```diff
@@ -45,3 +48,11 @@ def get(self, request, **kwargs):
             "row_limit": 3,
         }
         return render(request, "licences/licences.html", context)
+
+
+class Licence(TemplateView):
+    def get(self, request, pk):
+        licence, status_code = get_licence(request, pk)
+        if status_code != HTTPStatus.OK:
+            return Http404
```
andytaylorkainos, line 57: would prefer if you were able to show a custom error message

BenAndrew-Kainos: In this case this just means that the given ID doesn't exist, and so an error hasn't occurred but rather the page there looking for doesn't exist. hence the 404

BenAndrew-Kainos: Can still change if you wish

core/builtins/custom_tags.py
BenAndrew-Kainos: good point. done

core/builtins/custom_tags.py
BenAndrew-Kainos: I was very surprised to find this out. Timedelta doesn't support months :(

core/builtins/custom_tags.py
BenAndrew-Kainos: good point. done

core/builtins/custom_tags.py
BenAndrew-Kainos: done

licences/views.py
BenAndrew-Kainos: Or I could have two different error outcomes

BenAndrew-Kainos: Done

core/builtins/custom_tags.py
```diff
@@ -387,3 +412,8 @@ def abbreviate_string(string, length):
         return string
     else:
         return str(string)[:length] + "..."
+
+
+@register.filter()
+def get_usage_value(good):
+    return round((good["usage"] / good["licenced_quantity"]) * float(good["licenced_value"]), 2)
```
efahy, line 419: Any need to protect from a divide by zero?

BenAndrew-Kainos: Removed this function

PR #413, opened 2020-04-20
LT-1046: application-summary-page

rorysproule1:
- Reuse application.html to show a summary page on submission of an application

- Fix multiple bugs on the application page

- Ensure submission flow works for all application types

- Update e2es to work with new summary


applications/helpers/check_your_answers.py
```diff
@@ -418,3 +428,35 @@ def _is_application_export_type_temporary(application):
...
+
+
+def _convert_goods_categories(goods_categories):
+    return (", ".join([x["value"] for x in goods_categories]),)
+
+
+def get_licence_string(application):
+    application_type = application.case_type["sub_type"]["key"]
+    if application.case_type["reference"]["key"] == CaseTypes.SITL:
+        return applications.ApplicationPage.Summary.Licence.TRANSHIPMENT
```
BenAndrew-Kainos, line 448: Please change this to a dict somewhere with key as the casetype and value as the string

applications/services.py
```diff
@@ -358,7 +358,7 @@ def delete_goods_type_document(request, pk, good_pk):
 # Activity
 def get_activity(request, pk):
     data = get(request, CASES_URL + pk + ACTIVITY_URL)
-    return data.json()["activity"]
+    return data.json().get("activity")
```
BenAndrew-Kainos, line 361: Why was this changed out of interest?

applications/views/common.py
```diff
@@ -160,44 +167,73 @@ def dispatch(self, request, *args, **kwargs):
...
 
-        if "errors" in response:
-            return self.get(request, error=response["errors"]["text"][0], text=request.POST.get("text"), **kwargs)
+            response, _ = post_case_notes(request, self.case_id, request.POST)
 
-        return redirect(
-            reverse_lazy("applications:application", kwargs={"pk": self.application_id, "type": "case-notes"})
-        )
+            if "errors" in response:
+                return self.get(request, error=response["errors"]["text"][0], text=request.POST.get("text"), **kwargs)
```
BenAndrew-Kainos, line 232: This will only handle a single error, could you pass all errors in case multiple are returned

templates/applications/application.html
```diff
@@ -1,70 +1,110 @@
...
+			{% if type == "summary" %}
+				<dt class="govuk-summary-list__key">
+					{% if application.case_type.sub_type.key == 'hmrc' %}
+						{% lcs 'applications.ApplicationPage.Summary.QUERY' %}
+					{% else %}
+						{% lcs 'applications.ApplicationPage.Summary.LICENCE' %}
+					{% endif %}
+				</dt>
+				<dd class="govuk-summary-list__value" id="summary-list-reference-code">
+					{{ licence|default_na }}
```
BenAndrew-Kainos, line 83: is this trying to summaries the Licence?

templates/applications/application.html
```diff
@@ -117,73 +175,79 @@ <h1 class="govuk-heading-l">
...
+					<dd class="govuk-summary-list__value">
+						{% if application.case_officer.first_name %}
+							{{ application.case_officer.first_name }} {{ application.case_officer.last_name }}
+						{% else %}
+							{% lcs 'applications.ApplicationSummaryPage.NO_ASSIGNED_CASE_OFFICER' %}
+						{% endif %}
+					</dd>
+				</div>
+			{% endif %}
+			{% if type != "summary" %}
```
BenAndrew-Kainos, line 200: These ifs can be joined

ui_automation_tests/conftest.py
```diff
@@ -650,6 +650,11 @@ def agree_to_the_declaration(driver):  # noqa
     functions.click_submit(driver)
 
 
+@when("I agree to the application summary")
+def agree_to_the_summary(driver):  # noqa
+    functions.click_submit(driver)
```
BenAndrew-Kainos, line 655: We already have a When I click Submit step so I don't think you need to create a specific step for this

rorysproule1: Makes sense, removed new step and just used When i click continue

templates/applications/application.html
rorysproule1: joined them

applications/services.py
rorysproule1: Forgot to remove (was looking at a bug)

templates/applications/application.html
rorysproule1: renamed to application_type

applications/views/common.py
rorysproule1: Updated to handle multiple errors

applications/helpers/check_your_answers.py
rorysproule1: Made into dict

templates/applications/application.html
```diff
@@ -1,70 +1,110 @@
...
-			{% endif %}
-		</div>
+			</div>
+		{% endif %}
 	</div>
 
 	<div class="govuk-grid-row">
 		<dl class="govuk-summary-list govuk-summary-list--no-border govuk-grid-column-two-thirds">
-			{% if not application.case_type.sub_type.key == "hmrc" %}
+			{% if not application.case_type.sub_type.key == "hmrc" and type != "summary"%}
```
jfaracik, line 57: I'm not a fan of reusing `application.html`, could we have a discussion about this?

templates/applications/summary.html
```diff
@@ -0,0 +1,103 @@
+{% if not type or type == "summary"%}
```
jfaracik, line 1: Do we need `type == "summary"`

applications/forms/common.py
```diff
@@ -240,6 +240,7 @@ def declaration_form(application_id):
         ],
         default_button_name=strings.declaration.Declaration.BUTTON_TITLE,
         back_link=BackLink(
-            strings.BACK_TO_APPLICATION, reverse_lazy("applications:task_list", kwargs={"pk": application_id}),
+            strings.declaration.Declaration.BACK,
+            reverse_lazy("applications:application", kwargs={"pk": application_id, "type": "summary"}),
```
jfaracik, line 244: Is this going to the right place?

rorysproule1: Good spot!

rorysproule1: (Now redirecting to summary)

templates/applications/summary.html
rorysproule1: Removed. Now using new context value 'summary_page' to distinguish this conditional logic as it means 'type' no longer as two different uses

PR #424, opened 2020-04-29
feat: LT-1230 - Open Licence Type (Media)

iichr:
- Added a new question for OIEL licence type - media/crypto/military etc
- Modify the triage flow if Media is selected - auto set application to the temporary export type
- New page for static destinations which the user cannot edit/modify - includes all countries as set by default for the OIEL media type


templates/applications/goods-locations/static-all-destinations.html
```diff
@@ -0,0 +1,24 @@
...
+{% block back_link %}
+	<a href="{% url 'applications:task_list' application_id %}" id="back-link" class="govuk-back-link">Back to application overview</a>
+{% endblock %}
+
+{% block body %}
+	<div class="lite-app-bar">
+		<div class="lite-app-bar__content">
+			<h1 class="govuk-heading-l">
+			{% block title %}
+			Destinations
```
jfaracik, line 12: Indent pls

iichr: Indented, also added all the content strings

ui_automation_tests/conftest.py
```diff
@@ -173,6 +173,12 @@ def enter_application_name(driver, context):  # noqa
     functions.click_submit(driver)
 
 
+def choose_open_licence_category(driver, type_of_oiel, context):  # noqa
```
phlptt, line 176: dont think this needs to be in conftest

ui_automation_tests/features/submit_open_application.feature
```diff
@@ -94,3 +94,40 @@ Feature: I want to indicate the open licence I want
     When I go to exporter homepage
     And I click on applications
     Then I see submitted application
+
+  @LT_1230_open_application_export_licence_media_type
```
phlptt, line 98: need either @smoke or @regression here other wise it wont run via jenkins

ui_automation_tests/features/submit_open_application.feature
```diff
@@ -94,3 +94,40 @@ Feature: I want to indicate the open licence I want
...
+    And I save and continue on the summary page
+    And I click on the "route_of_goods" section
+    And I answer "Yes" for shipping air waybill or lading
+    And I click continue
+    And I click on the "temporary_export_details" section
+    And I provide details of why my export is temporary
+    And I answer "No" for whether the products remain under my direct control
+    And I enter the date "11", "05", "2027" when the products will return to the UK
+    And I save and continue on the summary page
+    When I click on the "countries" section
```
phlptt, line 121: Should be And rather than When

ui_automation_tests/conftest.py
iichr: Noted, will move when we make this more generic in the next story immediately after

ui_automation_tests/features/submit_open_application.feature
iichr: Adding

ui_automation_tests/features/submit_open_application.feature
iichr: Changed :)

PR #425, opened 2020-04-29
fix third parties

stephanieboyd:


applications/views/parties/third_parties.py
```diff
@@ -15,7 +15,7 @@ def get(self, request, **kwargs):
 
         context = {
             "application": application,
-            "third_parties": application["third_parties"],
+            "third_parties": get_third_parties(request, application_id),
```
andytaylorkainos, line 18: howcome we have to do this? (do we not return the third parties with the application anymore?)

stephanieboyd: On the application view, even deleted parties were shown. Checking the ultimate end users, it's uses an endpoint - so made them the same.

andytaylorkainos: oh - is there not a flag that indicates the user is deactivated?

PR #426, opened 2020-04-30
feat: LT-1413 Default Queue

andytaylorkainos:


applications/forms/common.py
```diff
@@ -232,7 +232,9 @@ def declaration_form(application_id):
                 title=strings.declaration.Declaration.RADIO_TITLE,
                 options=[
                     Option(
-                        "True", strings.declaration.Declaration.AGREE_TO_DECLARATION, test_id="agreed_to_declaration"
+                        "True",
```
BenAndrew-Kainos, line 235: Jan says we don't put quotes around the key if it's a boolean

BenAndrew-Kainos: also why did this file change? Is your black setup differently?

andytaylorkainos: data_attribute="agreed_to_declaration" is the change

jfaracik: Yeah, is converting to a string necessary?

PR #428, opened 2020-05-01
feat: LT-2061 and LT-1270 - OIEL Export Types (Cryptographic/Dealer/UK Continental Shelf)

iichr:
- Modify/create OIEL flows for the following 3 licence types: Cryptographic, Dealer and 🇬🇧 Continental Shelf
- Cryptographic OIEL's export type is automatically set to _permanent_
- Goods and countries are _non-editable_  on Crypto/Dealer OIELs, with both being restricted and pre-populated from a given limited set.
- Only countries on UK Continental Shelf OIELs are _non-editable_
- Cryptographic OIEL task list should only show the following 3 sections - Products, Destinations and Third Parties
- E2Es for the new flows

### DIFF
--- a/applications/constants.py
+++ b/applications/constants.py
@@ -34,7 +34,7 @@ class F680:
     }
 
 
-class OielLicenceTypes(Enum):
+class OielLicenceTypes:
     MEDIA = "media"
     CRYPTOGRAPHIC = "cryptographic"
     DEALER = "dealer"
--- a/applications/constants.py
+++ b/applications/constants.py
@@ -1,3 +1,6 @@
+from enum import Enum
+
+
 class F680:
     FIELDS = [
         "expedited",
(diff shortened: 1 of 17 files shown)


applications/constants.py
```diff
@@ -29,3 +29,8 @@ class F680:
         "expedited": "expedited_date",
         "locally_manufactured": "locally_manufactured_description",
     }
+
+
+class OielLicenceTypes:
```
a-bek, line 34: Can you use python enum:

```
class OielLicenceTypes(enum):
    MEDIA = "media"
    CRYPTOGRAPHIC = "cryptographic"
    DEALER = "dealer"
    UK_CONTINENTAL_SHELF = "uk_continental_sheld"

    def is_non_editable_good(cls, value):
         return value in [cls.MEDIA.value, cls.CRYTO.value]

    def is non_editable_country(cls, value):
        return value in [...]
```

iichr: Got it 🚀

iichr: Have a look, making it a subclass of Enum failed, but I created the class methods

PR #429, opened 2020-05-05
Performance: EUA list performance

BenAndrew-Kainos:


BenAndrew-Kainos:
> this creates an inconsistency between the dictionary present in EUAs and Standadrd/open applications - can you standardize it?

Yeah I'm implementing the change for applications in a following PR

PR #433, opened 2020-05-07
Performance: Goods endpoints

BenAndrew-Kainos:


goods/services.py
```diff
@@ -23,8 +23,8 @@ def get_goods(
     return data.json()
 
 
-def get_good(request, pk):
-    data = get(request, GOODS_URL + str(pk))
+def get_good(request, pk, full_detail=False):
+    data = get(request, GOODS_URL + str(pk) + f"?full_detail={full_detail}")
```
jfaracik, line 27: Could this be using the standard locals() way instead

PR #449, opened 2020-05-27
feat: LT-2520 - Specify location where records are held when creating a site

iichr:


organisation/sites/forms.py
```diff
@@ -69,8 +71,9 @@ def new_site_forms(request):
                 ],
                 default_button_name=generic.CONTINUE,
             ),
+            site_records_location(request),
             Form(
-                caption="Step 3 of 3",
```
Mark-Spiers-kainos, line 73: should this not be `4 of 4`

iichr: Good spot 👍

iichr: All changed!

PR #450, opened 2020-05-27
feat: LT-874 - Search open general licences

jfaracik:
**What's new**
* Added search for OGL flow

**Misc**
* Added LCS updates from internal (hidden text, pluralise)
* jQuery is now stored on our domain
* Cache control for assets - far better performance on environments (especially on Safari)
* `apple-touch-icons` now named correctly 😨


apply_for_a_licence/forms/triage_questions.py
```diff
@@ -103,7 +111,7 @@ def export_licence_questions(application_type, goodstype_category=None):
             ),
             *conditional(application_type == CaseTypes.OIEL, [goodstype_category_form()], []),
             *conditional(
-                goodstype_category not in ["media", "cryptographic"],
+                application_type != CaseTypes.OGEL,
```
iichr, line 114: Does this maintain the current behaviour, can't choose an export type for every application that's an OIEL. The e2es under create open application might be the fastest way to make sure :)

core/builtins/custom_tags.py
```diff
@@ -52,6 +57,22 @@ def get(object_to_search, nested_properties_list):
         return STRING_NOT_FOUND_ERROR
 
 
+@register.filter(name="lcsp")
+def pluralize_lcs(items, string):
```
iichr, line 61: 👌

PR #456, opened 2020-06-09
feat: Open Licence returns

BenAndrew-Kainos:


compliance/forms.py
```diff
@@ -0,0 +1,52 @@
...
+            ),
+            Form(
+                title=OpenReturnsForm.Year.TITLE,
+                description=OpenReturnsForm.Year.DESCRIPTION,
+                questions=[
+                    Select(
+                        title=OpenReturnsForm.Year.FIELD_TITLE,
+                        description=OpenReturnsForm.Year.FIELD_DESCRIPTION,
+                        name="year",
+                        options=get_years(),
```
jfaracik, line 36: If this is just two values (eg current year and previous year) could you use the radio buttons control instead :)

templates/core/hub.html
```diff
@@ -137,4 +137,26 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-5">
...
+		<h2 class="govuk-heading-m govuk-!-margin-bottom-5">{% lcs 'hub.Tiles.COMPLIANCE_HEADING' %}</h2>
+		<section class="govuk-form-group">
+			<div class="app-tiles">
+				{# Open licence returns tile #}
+				<div class="app-tile">
+					<a href="{% url 'compliance:open_licence_returns_list' %}" role="button" draggable="false" class="app-tile__link app-tile__heading" id="link-open-licence-returns">
+						{% lcs 'hub.Tiles.OPEN_LICENCE_RETURNS' %}
+					</a>
+					<p>{% lcs 'hub.Tiles.OpenLicenceReturns.DESCRIPTION' %}</p>
+					<a href="{% url 'compliance:add_open_licence_return' %}" class="govuk-link govuk-link--no-visited-state" role="button" draggable="false">
```
jfaracik, line 151: You won't need `role="button" draggable="false"`

compliance/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+class AnnualReturnsList(TemplateView):
+    def get(self, request, *args, **kwargs):
+        data = get_open_licence_returns(request)
+        return render(request, "compliance/open-licence-returns/list.html", {"open_licence_returns": data})
+
+
+class AnnualReturnsDownload(TemplateView):
+    def get(self, request, pk):
+        return get_open_licence_return_download(request, pk)
```
efahy, line 20: I notice we don't have the headings in the CSV when we download them and it looks a bit rubbish having an empty first row they are opened in Excel or whatever.  I think we should include them

compliance/forms.py
BenAndrew-Kainos: The design specifically wanted a select

compliance/views.py
BenAndrew-Kainos: There was a specific request when developing this story that we shouldn't store the CSV headers anywhere because we don't want to maintain the formatting rules. I agree it would look better with them but this is not possible unless we store them

PR #463, opened 2020-06-15
task: EUA e2e's

BenAndrew-Kainos:
Making EUA e2e's test properly & efficiently


ui_automation_tests/pages/end_user_advisory_page.py
```diff
@@ -1,29 +1,46 @@
...
+
+    def row_notifications(self):
+        return (
+            self.driver.find_element_by_css_selector(self.TABLE_ROW)
+            .find_element_by_css_selector(Shared.NOTIFICATION)
+            .text.split("\n")[0]
+        )
+
+    def click_row_copy(self):
+        self.driver.find_element_by_css_selector(self.TABLE_ROW).find_element_by_link_text("Copy").click()
```
andytaylorkainos, line 46: are you able to tie it in with the content used for the button? (not that its likely - but if the content changes, this test will fail)

BenAndrew-Kainos: One better, used an ID :)

PR #466, opened 2020-06-18
[Do Not Merge] Import products from SPIRE

richtier:


applications/views/goods.py
```diff
@@ -181,3 +240,43 @@ def get(self, request, **kwargs):
...
+        return kwargs
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(
+            application=get_application(self.request, self.kwargs["pk"]), draft_id=self.kwargs["pk"], **kwargs,
+        )
+
+        form = context["form"]
+        filters = form.cleaned_data if form.is_valid() else {}
+        respoonse = spire.client.list_licenses(
```
saruniitr, line 273: minor typo

applications/views/goods.py
```diff
@@ -169,6 +220,14 @@ def init(self, request, **kwargs):
         self.action = post_good_on_application
         self.success_url = reverse_lazy("applications:goods", kwargs={"pk": self.object_pk})
 
+    def get_initial_data(self):
+        for good in self.spire_licence["document_instance"]["document_data"]["goods_item_list"]["goods_item"]:
+            if good["part_no"] == self.request.GET["part_number"]:
```
saruniitr, line 225: is there a chance of multiple matches as we return on the first match?

richtier: possible, but (probably) unlikely. I would expect "quantity" to be used in that case. There is a quantity field that the user can use to specify if the product comes in multiples.

PR #469, opened 2020-06-22
bug: 1379 document upload

BenAndrew-Kainos:


applications/helpers/documents.py
```diff
@@ -0,0 +1,67 @@
...
+        return _extract_document_data(files, request.POST.get("description")), None
+    except UploadFailed:
+        if request.upload_handlers[0].content_length > MAX_UPLOAD_SIZE:
+            return None, strings.applications.AttachDocumentPage.FILE_TOO_LARGE
+        else:
+            return None, strings.applications.AttachDocumentPage.UPLOAD_FAILURE_ERROR
+
+
+def generate_file(result):
+    # Stream file upload
```
efahy, line 54: Download, not upload

applications/views/documents.py
```diff
@@ -50,40 +49,33 @@ def get(self, request, **kwargs):
         form = get_upload_page(request.path, draft_id)
         return form_page(request, form, extra_data={"draft_id": draft_id})
 
+    @staticmethod
+    def _perform_upload(request, draft_id, data, kwargs):
```
efahy, line 53: I think `_perform_upload` is not a good name - the actual upload happens in `handle_document_upload`.  What about `_save_document_metadata` for this or `_save_document_on_application` or something similar for this method?

applications/views/documents.py
```diff
@@ -50,40 +49,33 @@ def get(self, request, **kwargs):
...
+            _, status_code = action(request, draft_id, data)
+            if status_code == HTTPStatus.CREATED:
+                return get_homepage(request, draft_id)
+        else:
+            _, status_code = action(request, draft_id, kwargs["obj_pk"], data)
+            if status_code == HTTPStatus.CREATED:
+                return get_homepage(request, draft_id, kwargs["obj_pk"])
+
+        # If status isn't CREATED, show error page
+        return error_page(request, strings.applications.AttachDocumentPage.UPLOAD_FAILURE_ERROR)
```
efahy, line 65: If this operation isn't successful, shouldn't we remove the file we uploaded to S3, otherwise it'll sit there orphaned forever?

PR #479, opened 2020-06-25
feat: LT-2632 - I want to see which OGLs I have already registered for when searching for new ones

jfaracik:
image

### DIFF
--- a/core/services.py
+++ b/core/services.py
@@ -22,6 +22,7 @@ from conf.constants import (
     LICENCES_OPEN_GENERAL_POST_URL,
 )
 from core.helpers import convert_parameters_to_query_params, convert_value_to_query_param
+from lite_content.lite_exporter_frontend.applications import OpenGeneralLicenceQuestions
 from lite_content.lite_exporter_frontend.generic import Document
 from lite_forms.components import Option, TextArea
 from lite_forms.generators import error_page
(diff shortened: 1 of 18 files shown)


core/services.py
```diff
@@ -304,9 +305,13 @@ def get_open_general_licences(
     if convert_to_options:
         return [
             Option(
-                ogl["id"],
-                ogl["case_type"]["reference"]["value"] + " (" + ogl["name"] + ")",
+                key=ogl["id"],
+                value=ogl["case_type"]["reference"]["value"] + " (" + ogl["name"] + ")",
```
BenAndrew-Kainos, line 309: please use an f string instead

core/services.py
```diff
@@ -304,9 +305,13 @@ def get_open_general_licences(
...
+                tag=OpenGeneralLicenceQuestions.OpenGeneralLicences.ALREADY_REGISTERED
+                if bool(len(ogl["registrations"]))
```
BenAndrew-Kainos, line 311: You don't need to wrap this in bool()

core/services.py
```diff
@@ -304,9 +305,13 @@ def get_open_general_licences(
...
+                else None,
                 more_information=ogl["description"],
+                disabled=bool(len(ogl["registrations"])),
```
BenAndrew-Kainos, line 314: same here

templates/components/ogl-step-list.html
```diff
@@ -1,6 +1,52 @@
...
+					{% for country in data.countries %}
+						{{ country.name }}{% if not forloop.last %}, {% endif %}
+					{% endfor %}
+				</dd>
+			</div>
+		</dl>
+	</li>
+	<li>
+		<p>{% lcs 'applications.OpenGeneralLicenceQuestions.OpenGeneralLicenceDetail.Points.SECOND_POINT' %}</p>
+		<span>{{ data.url|linkify }}</span>
```
BenAndrew-Kainos, line 35: Instead of linkify, could we just use markdown (which supports markdown links -> html)

core/services.py
jfaracik: Can't access dict keys in an F string :/

templates/components/ogl-step-list.html
jfaracik: We just want to print the URL so linkify is appropriate here :)

PR #488, opened 2020-07-10
Add sanity checks for organisation

saruniitr:
Add sanity check for Organisation
- confirming an organisation where there is nothing in review
- get list of application without registering an organisation


saruniitr:
yes, on it

PR #495, opened 2020-07-20
bug: LT-2794 - Disallow UK end users on SITLs

iichr:
- United Kingdom shouldn't be an option to select for end users on standard transhipments

### DIFF
--- a/applications/forms/parties.py
+++ b/applications/forms/parties.py
@@ -1,5 +1,6 @@
 from django.urls import reverse_lazy
 
+from conf.constants import CaseTypes
 from core.services import get_countries
 from lite_content.lite_exporter_frontend import strings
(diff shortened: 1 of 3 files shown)


applications/forms/parties.py
```diff
@@ -104,6 +108,11 @@ def new_party_form_group(request, application, strings, back_url, clearance_opti
     if clearance_options:
         forms.extend(clearance_level_forms(clearance_options, strings.BUTTON))
 
-    forms.append(party_address_form(request, strings.ADDRESS_FORM_TITLE, strings.SUBMIT_BUTTON))
+    # Exclude the UK if end user on standard transhipment
+    is_end_user = "/end-user" in request.path
```
BenAndrew-Kainos, line 112: Could we change this to instead pass `is_end_user` as an argument. We shouldn't have code that relies on the URL path

